PR #34, opened 2026-04-13
fix: NaN occurs_loss when every sample in a batch is censored

### LINKED ISSUE #30: NaN `occurs_loss` (and therefore total loss + gradients) when all samples in a batch are censored
## Summary

In `EveryQueryModel._forward` (`src/every_query/model.py:655`), `occurs_loss` is computed as a BCE-with-logits loss masked by `~batch.censor`:

```python
censor_logits = self.censor_mlp(query_embed)
censor_loss = self._get_loss(censor_logits, batch.censor, mask=None)

# if future is censored then don't apply loss on whether query occurs
occurs_logits = self.occurs_mlp(query_embed)
occurs_loss = self._get_loss(occurs_logits, batch.occurs, mask=~batch.censor)

loss = censor_loss + occurs_loss
```

When every sample in the batch is censored, `~batch.censor` is all-False, so `_get_loss` applies the mask and hands `BCEWithLogitsLoss` two empty tensors of shape `(0, 1)`. With the default `reduction='mean'`, `BCEWithLogitsLoss(empty, empty) == nan` (0/0).

Verified locally:

```python
>>> import torch
>>> crit = torch.nn.BCEWithLogitsLoss()
>>> loss = crit(torch.zeros((0, 1), requires_grad=True), torch.zeros((0, 1)))
>>> loss
tensor(nan, grad_fn=<BinaryCrossEntropyWithLogitsBackward0>)
>>> torch.isnan(loss).item()
True
```

That NaN then propagates into `loss = censor_loss + occurs_loss`, so the total loss returned by the forward pass is NaN, and `loss.backward()` in `training_step` destroys the optimizer state with NaN gradients.

## When does this trigger in practice?

A batch is "all censored" exactly when every prediction_time in the batch has less than `duration` days of future data. Because:

- Censoring rates grow monotonically with the requested duration (labels are generated for durations 30..365 days).
- `min_context_per_subject` only bounds past context, not future horizon.
- With imbalanced shards or long-duration tasks (e.g., 365 days), per-batch censoring rates of 50–90% are realistic.

So with e.g. 80% censoring and `batch_size=8`, roughly 1 in 6 batches will be fully censored — enough to blow up a training run. The bug is silent: it manifests as NaN loss, the optimizer step NaN-poisons the model, and every subsequent forward produces NaN logits. Lightning's default `detect_anomaly=False` does not surface it.

`_log_metrics` (`lightning_module.py:253-264`) already guards the metrics branch with `if preds.numel() > 0 and target.unique().numel() == 2`, but the loss path has no equivalent guard. That's the inconsistency that makes this a bug rather than a known limitation.

## Severity

**High / critical.** Silent NaN training. No runtime error, no warning, just a corrupted model.

## Proposed fix

Return a zero (detached-from-input but device/dtype-correct) loss when the mask is empty:

```python
def _forward(self, batch: EveryQueryBatch):
    ...
    censor_logits = self.censor_mlp(query_embed)
    censor_loss = self._get_loss(censor_logits, batch.censor, mask=None)

    occurs_logits = self.occurs_mlp(query_embed)
    uncensored_mask = ~batch.censor
    if uncensored_mask.any():
        occurs_loss = self._get_loss(occurs_logits, batch.occurs, mask=uncensored_mask)
    else:
        occurs_loss = occurs_logits.sum() * 0.0  # zero, but differentiable / on the right device+dtype
    loss = censor_loss + occurs_loss
    ...
```

Using `occurs_logits.sum() * 0.0` (or equivalently `torch.zeros((), device=..., dtype=...)`) keeps the loss as a proper 0-dim tensor with the right device/dtype and maintains autograd graph connectivity (no dangling parameter-less loss).

Alternatively, guard inside `_get_loss` itself so any future masked loss head benefits:

```python
def _get_loss(self, logits, target, mask=None):
    target = target.float().unsqueeze(1)
    if mask is not None:
        logits = logits[mask]
        target = target[mask]
        if logits.numel() == 0:
            return logits.sum() * 0.0
    return self.criterion(logits, target)
```

## Test

A regression test should construct a batch where `batch.censor` is all-True and assert:
1. `_forward(batch)` returns a finite total loss.
2. `outputs.occurs_loss` is finite (ideally exactly zero).
3. Calling `loss.backward()` does not introduce NaNs into any `occurs_mlp` parameter.

mmcdermott:
## Summary

Fixes #30.

- `model.py::_forward` passes `mask=~batch.censor` into `_get_loss` for the occurs head. When every sample in a batch is censored that mask is all-False, the masked logits/target tensors become shape `(0, 1)`, and `BCEWithLogitsLoss(reduction='mean')` returns NaN on an empty input. That NaN flows into the total loss and into `backward()`, silently poisoning optimizer state.
- This PR guards `_get_loss` to return a differentiable zero (`logits.sum() * 0.0`) when the masked batch is empty, which keeps the autograd graph connected and preserves the correct device/dtype.

## Structure

Two commits, as requested:

1. **`test:` failing regression** — adds `TestAllCensoredBatchProducesFiniteLoss` with two cases: (a) forward on an all-censored batch must yield finite `occurs_loss` and finite total loss, (b) `loss.backward()` on the same batch must not produce NaN gradients on any parameter. On the pre-fix `model.py` both cases fail with `tensor(nan)`.
2. **`fix:`** — guards the empty-mask branch in `_get_loss` and adds a short doctest covering the zero-loss path (return value, autograd connectivity).

## Test plan

- [x] `uv run pytest tests/test_model_logic.py::TestAllCensoredBatchProducesFiniteLoss` — both new tests pass post-fix.
- [x] `uv run pytest tests/test_model_logic.py tests/test_lightning_logic.py tests/test_dataset_logic.py src/every_query/model.py --doctest-modules` — 45 passed, no regressions.
- [ ] Spot-check a real training run with a high-censoring-rate config (e.g. 365-day duration on a small shard) to confirm training no longer NaNs.


tests/test_model_logic.py
```diff
@@ -124,6 +124,56 @@ def test_flipping_censored_occurs_label_preserves_occurs_loss(self, demo_model,
         )
 
 
+class TestAllCensoredBatchProducesFiniteLoss:
```
mmcdermott, line 127: This is redundant with the doctest, no?

mmcdermott: I think they're at different layers — would push back on dropping the whole class, but happy to trim.

- **Doctest** (in `_get_loss`): unit-tests the *fix site* — `_get_loss` with an all-False mask returns `0.0` with `requires_grad=True`. It calls `_get_loss` directly and doesn't exercise `_forward` or the mask construction.
- **`test_all_censored_batch_backward_does_not_nan_params`**: exercises the *bug site* from #30. The bug is `_forward` building `mask=~batch.censor`, handing it to `_get_loss`, and the NaN then flowing through `loss = censor_loss + occurs_loss` into `backward()` and poisoning optimizer state. The doctest's `requires_grad=True` check is necessary-but-not-sufficient for that — e.g., if someone later changed `_forward` to build the mask differently, or added a NaN-producing op on the all-censored path, the doctest would still pass while the real regression re-surfaced. This test is the end-to-end guard and I'd keep it.
- **`test_all_censored_batch_total_loss_is_finite`**: this one is the weakest of the three. If `backward()` produces no NaN grads, forward must have been finite, so it's largely subsumed by the backward test. Its only extra value is the per-field assertion (`censor_loss` vs `occurs_loss` individually finite) for a cleaner failure message. Happy to drop it if you'd prefer — say the word and I'll remove it in a follow-up commit.

PR #44, opened 2026-04-15
Fix gsutil rsync hanging on macOS

payalchandak:
## Summary
- Disables `parallel_process_count` on macOS to work around [CPython issue 33725](https://bugs.python.org/issue33725), which causes `gsutil -m rsync` to hang indefinitely during "Building synchronization state"
- Multithreading is still used for download speed; only multiprocessing (which deadlocks on macOS) is disabled

## Test plan
- [x] Verified `python scripts/sync.py download` completes successfully on macOS after the fix (previously hung on the `outputs` mapping)


mmcdermott:
If we remove this code per #67 this won't be needed.

Issue #54, opened 2026-04-16
Refactor EveryQuery for external usability: script endpoints, documentation, and clear pipeline stages


mmcdermott:
## Summary

EveryQuery needs a major refactor to make it usable by external parties — modeled after [MEDS_EIC_AR](https://github.com/mmcdermott/MEDS_EIC_AR). Currently, the project has no `[project.scripts]` console entry points, a minimal README (25 lines of mostly-stale content), and no clear documentation of the end-to-end pipeline. After `pip install EveryQuery`, users should be able to run the full pipeline (preprocess → prepare tasks → train → evaluate/infer) via named CLI commands with Hydra configuration.

## Current State

| Concern | MEDS_EIC_AR (reference) | EveryQuery (today) |
|---------|------------------------|--------------------|
| Console scripts | 3 (`MEICAR_process_data`, `MEICAR_pretrain`, `MEICAR_generate_trajectories`) | 0 — all invoked via `python src/every_query/train.py` etc. |
| README | ~300 lines: install, 3-step usage, config docs, output format | ~25 lines: stale .env snippet, references wrong commands |
| Config discovery | `importlib.resources.files(...)` for package-relative configs | Relative `config_path=""` — breaks after pip install |
| Preprocessing | In-package `preprocessing/` submodule with Hydra entry point | **Not in this repo at all** — README references `MEICAR_process_data` from the other project |
| Task preparation | N/A (AR model doesn't need separate task labels) | `tasks.py` and `sample_tasks.py` exist but have no console scripts |
| Evaluation | Integrated into `generate_trajectories` | Multi-stage eval suite (`gen_index_times` → `gen_task` → `eval` → `select_model`) — no console scripts |

## Target Architecture

After this refactor, the following CLI commands should be available after `pip install EveryQuery`:

```bash
# 1. Preprocess raw MEDS data into tensorized format
EQ_process_data input_dir="$RAW" intermediate_dir="$INTERMEDIATE" output_dir="$FINAL_DATA_DIR"

# 2. Generate task labels for pre-training
EQ_generate_tasks data_dir="$FINAL_DATA_DIR" task_dir="$TASK_DIR" ...

# 3. Train the model
EQ_train datamodule.config.tensorized_cohort_dir="$FINAL_DATA_DIR" \
    query.task_dir="$TASK_DIR" output_dir="$MODEL_DIR"

# 4. Evaluate / run inference
EQ_evaluate model_run_dirs=["$MODEL_DIR"] \
[...]

PR #61, opened 2026-04-16
feat: console script endpoints and importlib.resources config paths

### LINKED ISSUE #60: Refactor config paths for installable package (importlib.resources)
Parent: #54

## Goal

Refactor all Hydra `config_path` arguments to use `importlib.resources.files()` so configs are correctly resolved when the package is pip-installed (not just run from the source tree).

## Problem

Currently, all Hydra entry points use relative `config_path` strings:

```python
# train.py
@hydra.main(version_base="1.3", config_path="", config_name="config.yaml")

# eval.py
@hydra.main(version_base="1.3", config_path="./eval_suite/conf", config_name="eval_config.yaml")

# sample_tasks.py
@hydra.main(version_base=None, config_path=".", config_name="sample_tasks_config")

# tasks.py
@hydra.main(version_base=None, config_path=".", config_name="tasks_config")

# eval_suite/gen_index_times.py
@hydra.main(config_path="./conf", config_name="gen_index_times_config", version_base=None)

# eval_suite/gen_task.py
@hydra.main(config_path="./conf", config_name="gen_tasks_config", version_base=None)

# eval_suite/select_model.py
@hydra.main(version_base="1.3", config_path="./conf", config_name="select_model_config.yaml")
```

These resolve relative to the Python source file, which works during development but breaks when the package is installed into a site-packages directory (the YAML files are there too, but the relative path resolution can be fragile and framework-dependent).

## Solution

Follow [MEDS_EIC_AR's pattern](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/__main__.py):

```python
from importlib.resources import files

CONFIGS = files("every_query")

@hydra.main(version_base="1.3", config_path=str(CONFIGS), config_name="config")
def main(cfg: DictConfig) -> float | None:
    ...
```

## Files to Modify

| File | Current `config_path` | Proposed |
|------|----------------------|----------|
| `train.py` | `""` | `str(files("every_query"))` |
| `eval.py` | `"./eval_suite/conf"` | `str(files("every_query") / "eval_suite" / "conf")` |
| `sample_tasks.py` | `"."` | `str(files("every_query"))` |
| `tasks.py` | `"."` | `str(files("every_query"))` |
| `eval_suite/gen_index_times.py` | `"./conf"` | `str(files("every_query") / "eval_suite" / "conf")` |
| `eval_suite/gen_task.py` | `"./conf"` | `str(files("every_query") / "eval_suite" / "conf")` |
| `eval_suite/select_model.py` | `"./conf"` | `str(files("every_query") / "eval_suite" / "conf")` |
| `process_composite/process_composite.py` | TBD | `str(files("every_query") / "process_composite")` |

## Additional Considerations

1. **`pyproject.toml` already includes YAML**: `[tool.setuptools.package-data] every_query = ["**/*.yaml"]` — this is correct and ensures YAML files are included in the wheel.

2. **Config store entries**: If any configs use Hydra's `ConfigStore` (like MEDS_EIC_AR does for `MEDSTorchDataConfig`), those registrations need to happen before Hydra's `@hydra.main` decorator runs. Verify this for the datamodule config.

3. **Defaults lists**: Check that `defaults:` entries in YAML configs (like `train_codes: 10000_ID__8db2be6fadf8`) still resolve correctly with the new config path.

4. **Test impact**: Tests that call Hydra functions directly may need adjustment if they relied on the relative path behavior. Run full test suite after changes.

## Acceptance Criteria

- [ ] All Hydra entry points use `importlib.resources.files`
- [ ] Package installs and runs correctly via `pip install .`
- [ ] All existing tests pass
- [ ] YAML configs are included in the built wheel
- [ ] Verified with `pip install . && EQ_train --help` (or equivalent)

## This is a prerequisite for all other script endpoint issues

This should be done first (or concurrently with the first script endpoint), as all console scripts depend on correct config resolution.

mmcdermott:
## Summary

- Refactor all Hydra `config_path` arguments across 11 files to use `importlib.resources.files()` for package-relative config resolution, ensuring configs are found correctly after `pip install`
- Add 7 console script entry points to `pyproject.toml`: `EQ_train`, `EQ_evaluate`, `EQ_generate_tasks`, `EQ_generate_tasks_exhaustive`, `EQ_gen_eval_index`, `EQ_gen_eval_tasks`, `EQ_select_model`
- All 203 tests pass, all pre-commit hooks pass, ruff clean

Closes #60
Part of #54

## Test plan

- [x] All 203 existing tests pass
- [x] All pre-commit hooks pass (ruff format, ruff lint, codespell, etc.)
- [x] Console scripts install and resolve configs correctly (`EQ_train --help`, `EQ_generate_tasks --help`, etc.)
- [x] Config search paths point to the correct package-relative directories
- [ ] CI tests pass

### DIFF
--- a/tests/test_cli_smoke.py
+++ b/tests/test_cli_smoke.py
@@ -1,0 +1,100 @@
+"""Subprocess smoke tests for the ``EQ_*`` console script entry points.
+
+Each endpoint is invoked via its installed console script name (the contract
+introduced in PR #61) with ``--help``.  A successful exit proves four things
+at once:
+
+1. The ``[project.scripts]`` entry in ``pyproject.toml`` resolved and the
+   installed shim is runnable.
+2. Hydra accepted the ``config_path=`` argument (this PR's switch from
+   fragile relative paths to ``importlib.resources.files()``).
+3. Hydra could locate and load the named config file under that path — if
+   the package-data glob or the resource resolution regressed, ``--help``
+   would fail with ``MissingConfigException``.
+4. Module-level imports don't blow up in a fresh interpreter.
+
+Three endpoints (``EQ_train``, ``EQ_evaluate``, ``EQ_gen_eval_tasks``)
+compose a ``{train,eval}_codes`` default group whose canonical file is
+generated out-of-band (see ``src/every_query/sample_codes/``) and is not
+checked in.  For those we point Hydra at a throwaway ``--config-dir`` that
+supplies an empty-codes smoke variant, which preserves the rest of the
+compose path.
+
+Subprocess coverage follows the pytest-cov recipe:
+https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html
+"""
+
+from __future__ import annotations
+
+import os
+import subprocess
+import sys
+from pathlib import Path
+
+import pytest
+
+_VENV_BIN = str(Path(sys.executable).parent)
+_REPO_ROOT = Path(__file__).resolve().parent.parent
+
+# (console_script, extra_args) — extras inject a smoke code-group for configs
+# whose defaults pull an out-of-tree YAML.
+_ENTRYPOINTS: list[tuple[str, list[str]]] = [
+    ("EQ_train", ["train_codes=smoke"]),
+    ("EQ_evaluate", ["eval_codes=smoke"]),
+    ("EQ_generate_tasks", []),
+    ("EQ_generate_tasks_exhaustive", []),
+    ("EQ_gen_eval_index", []),
+    ("EQ_gen_eval_tasks", ["eval_codes=smoke"]),
+    ("EQ_select_model", []),
+]
+
+
+@pytest.fixture(scope="module")
+def smoke_config_dir(tmp_path_factory) -> Path:
+    """Temp Hydra search dir supplying empty ``train_codes`` / ``eval_codes``."""
+    d = tmp_path_factory.mktemp("eq_smoke_cfg")
+    (d / "train_codes").mkdir()
+    (d / "train_codes" / "smoke.yaml").write_text("codes: []\n")
+    (d / "eval_codes").mkdir()
+    (d / "eval_codes" / "smoke.yaml").write_text("id: []\nood: []\n")
+    return d
+
+
+@pytest.fixture(scope="module")
+def cli_env(tmp_path_factory) -> dict[str, str]:
+    """Subprocess env with venv ``PATH`` and pytest-cov subprocess hook.
+
+    Writes an ephemeral ``sitecustomize.py`` that calls
+    ``coverage.process_startup()`` and points the child at it via
+    ``PYTHONPATH`` + ``COVERAGE_PROCESS_START``.  This is a no-op when
+    coverage is not installed or not active; when it is,
+    ``[tool.coverage.run] parallel = true`` in ``pyproject.toml`` makes
+    the child-process data files merge cleanly with the parent's.
+    """
+    env = os.environ.copy()
+    env["PATH"] = _VENV_BIN + os.pathsep + env.get("PATH", "")
+
+    sitecustom = tmp_path_factory.mktemp("eq_cov_sitecustomize")
+    (sitecustom / "sitecustomize.py").write_text(
+        "try:\n    import coverage\n    coverage.process_startup()\nexcept Exception:\n    pass\n",
+    )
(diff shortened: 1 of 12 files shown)


mmcdermott:
Can you add some simple smoke tests that run these new CLI endpoints through subprocess and validates that, on plausible inputs, they don't crash and/or return usable signals? That will make sure they are runnable, at least. 

If you do this, you'll need to add the subprocess patch to pytest-cov to make sure they get reflected in codecov: https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html

mmcdermott:
Done in b5e4b3c — `tests/test_cli_smoke.py` parametrizes all 7 `EQ_*` scripts through `subprocess.run` with `--help` and asserts exit 0 + Hydra help signature.  `--help` drives Hydra's full compose, so it exercises the entry-point shim, `importlib.resources` config-path resolution, and module imports in one shot — the exact regression surface this PR touches.

Three configs (`config.yaml`, `eval_config.yaml`, `gen_tasks_config.yaml`) pull a `{train,eval}_codes/<hash>.yaml` default that's generated out of tree by `src/every_query/sample_codes/` and isn't checked in, so the test injects an empty smoke variant via `--config-dir`.  Noticed in passing: `EQ_train --help`, `EQ_evaluate --help`, `EQ_gen_eval_tasks --help` all fail from a fresh clone on `main` for this reason — separate cleanup, happy to open an issue if you want.

Subprocess coverage wired per the pytest-cov recipe:
- `[tool.coverage.run] parallel = true` + `source_pkgs = ["every_query"]` in `pyproject.toml`
- fixture writes an ephemeral `sitecustomize.py` that calls `coverage.process_startup()` and points the child at it via `PYTHONPATH` + `COVERAGE_PROCESS_START`

`uv run pytest tests/test_cli_smoke.py --cov=every_query` now reports 17-31% coverage on modules only reached via child processes (`train.py`, `eval.py`, `tasks.py`, `sample_tasks.py`, `eval_suite/*`), confirming the subprocess attribution works end-to-end.  Full suite green: 230 passed.

tests/test_cli_smoke.py
```diff
@@ -0,0 +1,100 @@
...
+    d = tmp_path_factory.mktemp("eq_smoke_cfg")
+    (d / "train_codes").mkdir()
+    (d / "train_codes" / "smoke.yaml").write_text("codes: []\n")
+    (d / "eval_codes").mkdir()
+    (d / "eval_codes" / "smoke.yaml").write_text("id: []\nood: []\n")
+    return d
+
+
+@pytest.fixture(scope="module")
+def cli_env(tmp_path_factory) -> dict[str, str]:
```
mmcdermott: You don't need to do any of this. There is a much easier way for coverage. See https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html

tests/test_cli_smoke.py
```diff
@@ -0,0 +1,100 @@
...
+    """``<script> --help`` exits 0 and produces Hydra's help signature."""
+    cmd = [script, f"--config-dir={smoke_config_dir}", *extra_args, "--help"]
+    result = subprocess.run(cmd, capture_output=True, text=True, env=cli_env, timeout=60)
+
+    combined = result.stdout + result.stderr
+    assert result.returncode == 0, (
+        f"{script} --help failed (rc={result.returncode})\n"
+        f"cmd: {cmd}\nstdout:\n{result.stdout}\nstderr:\n{result.stderr}"
+    )
+    assert "Powered by Hydra" in combined, (
```
mmcdermott: This assertion is not necessary. The existence of the CLI endpoint is what is tested, not that it uses hydra.

tests/test_cli_smoke.py
mmcdermott: You're right, that was overkill. Fixed in b53493f — replaced the whole env-wiring dance with `[tool.coverage.run] patch = ["subprocess"]` in `pyproject.toml`. Coverage numbers on subprocess-only modules are unchanged (17-31% across train/eval/tasks/sample_tasks/eval_suite/*).

tests/test_cli_smoke.py
mmcdermott: Fair — dropped in b53493f. The rc==0 on `--help` is the contract the test cares about.

Issue #67, opened 2026-04-17
`upload_models.py` is package-internal but only invoked via a Columbia-cluster SLURM wrapper


mmcdermott:
## What I found

`src/every_query/upload_models.py` (122 lines, a GCS-sync utility) has no Python imports anywhere in the repo. Its only invocation path is `scripts/upload_models.sh`, a SLURM submit script that's hard-wired to a Columbia cluster account:

```bash
# scripts/upload_models.sh
#SBATCH --mail-user=gbk2114@cumc.columbia.edu
...
uv run python src/every_query/upload_models.py --runs-file ...
```

So from the installed-wheel perspective it's dead weight, and from an external-user perspective it's unrunnable without Columbia cluster setup.

## Options

1. **Promote** — add \`EQ_upload_models\` to \`[project.scripts]\`, make the config (bucket, path layout) user-provided via Hydra/argparse rather than \`scripts/sync_config.yaml\`. Document as \"optional: push trained checkpoints to GCS\". Valuable if other groups would also push to a bucket.
2. **Demote** — move \`upload_models.py\` out of \`src/every_query/\` into \`scripts/\` alongside \`upload_models.sh\`, clearly marking it as a private operator tool. No longer ships in the wheel. Keeps the public package lean.
3. **Delete** — if it's only used by one group on one cluster and nobody else needs it.

I'd lean toward **demote** unless there's concrete plan to support external uploads. The SLURM wrapper + `scripts/sync_config.yaml` already imply private-operator usage.

cc @payalchandak (author, 3b93992) — which direction?

mmcdermott:
We should drop this. This is the kind of script that should be local, or in an internal-only companion repo.

Issue #64, opened 2026-04-17
Config defaults reference gitignored `{train,eval}_codes/*.yaml` — `EQ_train --help` fails on a fresh clone


mmcdermott:
## What I found

Three configs pull a default code-group file that is (a) generated out-of-band by a script with hardcoded MIMIC paths and (b) explicitly \`.gitignore\`d, so a fresh clone cannot compose them:

| Config | Default group entry | Provider file |
|---|---|---|
| \`src/every_query/config.yaml\` | \`train_codes: 10000_ID__8db2be6fadf8\` | \`src/every_query/train_codes/10000_ID__8db2be6fadf8.yaml\` (gitignored) |
| \`src/every_query/eval_suite/conf/eval_config.yaml\` | \`eval_codes: 10000_ID__8db2be6fadf8\` | \`src/every_query/eval_suite/conf/eval_codes/10000_ID__8db2be6fadf8.yaml\` (gitignored) |
| \`src/every_query/eval_suite/conf/gen_tasks_config.yaml\` | \`eval_codes: 10000_ID__8db2be6fadf8\` | (same as above) |

Confirmed in \`.gitignore:184,188\`:
\`\`\`
src/every_query/eval_suite/conf/eval_codes/
src/every_query/train_codes/
\`\`\`

The generator scripts under \`src/every_query/sample_codes/\` hard-code a MIMIC path:
\`\`\`python
# src/every_query/sample_codes/sample_train_codes.py:10
PARQUET_PATH = "/users/gbk2114/data/MIMIC_MEDS/MEDS_cohort/processed/metadata/codes.parquet"
\`\`\`

so they can't run for anyone outside that cluster either.

## Concrete symptom

On a clean clone with no prior setup:
\`\`\`console
$ EQ_train --help
hydra.errors.MissingConfigException: In 'config.yaml': Could not find 'train_codes/10000_ID__8db2be6fadf8'
\`\`\`

Same for \`EQ_evaluate --help\` and \`EQ_gen_eval_tasks --help\`. The new smoke tests in \`tests/test_cli_smoke.py\` work around this by injecting an empty \`{train,eval}_codes/smoke.yaml\` via \`--config-dir\`.

## Proposed solutions (pick one)

[...]

PR #76, opened 2026-04-17
refactor: remove tasks.py + collate_tasks (closes #53)

### LINKED ISSUE #53: Refactor: remove dead tasks.py code superseded by sample_tasks.py
## Context

PR #41 added `sample_tasks.py` as a sampling-first replacement for the exhaustive per-(shard x duration x code) label generation in `tasks.py`. As noted in that PR's "Explicitly out of scope" section, the old code was left in place until the sampler was validated in a real pre-training run.

Now that #41 has landed, the legacy code path can be removed. This issue tracks that cleanup.

### `src/every_query/tasks.py` (entire module)
All functions are superseded by `sample_tasks.py` equivalents:
- `read_event_shard` -> `sample_tasks._read_event_shard` (with dtype normalization)
- `read_query_codes` -> `sample_tasks._read_query_codes` (with deterministic sort)
- `compute_base_prediction_times` -> `sample_tasks.sample_contexts`
- `derive_censor_for_duration` -> folded into `sample_tasks.evaluate_index_df`
- `precompute_min_deltas_wide` -> replaced by single-pass `join_asof` in `evaluate_index_df`
- `build_task_for_duration` -> replaced by `evaluate_index_df`
- `sample_durations` -> `sample_tasks.sample_tasks` (draws durations per-task from log-uniform)
- `main` (Hydra entry point) -> `sample_tasks.main`

Also remove `tasks_config.yaml`.

### `src/every_query/tasks_reference.py` (entire module)
Reference/oracle implementation (`compute_censor_dataframe`, `build_task_label_matrix`) used only in tests as a cross-check. Once `tasks.py` is removed, the reference is only needed by `test_sample_tasks.py` (which already has its own cross-check). Evaluate whether to keep it as a test-only oracle or remove it entirely.

### `src/every_query/train.py::collate_tasks` and `_collate_shard`
The collation step that reshapes the old wide-format per-duration task parquets into the long-format the dataloader expects. `sample_tasks.py` outputs the long format directly, so collation is unnecessary. `train.main` currently calls `collate_tasks` on every run (line 349); this needs to be replaced with a path to the sampler output.

### Test files
- `tests/test_tasks.py` -- tests for the old `tasks.py` functions, all cross-checked against `tasks_reference`. Remove or merge any remaining coverage gaps into `test_sample_tasks.py`.
- `tests/test_collate_tasks.py` -- tests for `collate_tasks` and `_collate_shard`. Remove entirely.
- `conftest.py::task_parquet_dir` fixture (and any other fixtures that exist solely to support the old code path).

## What to preserve

- `tasks_reference.py` may be worth keeping as a test-only oracle if `test_sample_tasks.py` still cross-checks against it. If so, move it to `tests/` or gate it behind a test-only import.
- Any test coverage from `test_tasks.py` that exercises edge cases not yet covered by `test_sample_tasks.py` should be ported before deletion.

## Depends on

- #41 merged
- `train.py` entry point updated to consume sampler output (may be done in this PR or a separate one)

### LINKED ISSUE #63: Move `tasks_reference.py` out of the production package into tests
## What I found

\`src/every_query/tasks_reference.py\` sits at the top of the production package, but its module docstring says (verbatim):

> These exist solely for regression testing: tests compare the output of the new precompute_min_deltas_wide + build_task_for_duration pipeline against these originals. **Do not import these in production code.**

Inbound imports (all under \`tests/\`):

- \`tests/test_sample_tasks.py:26\` — \`from every_query import tasks_reference\`
- \`tests/test_tasks.py:8\` — \`from every_query import tasks, tasks_reference\`

No production module imports it.

## Proposed change

Move the file under \`tests/\` (e.g. \`tests/fixtures/tasks_reference.py\` or \`tests/_tasks_reference.py\`) and update the two test imports. Keeps it out of the installable wheel and removes the \"don't import me\" footgun from the public namespace.

## Related

Possibly overlaps with #53 (remove dead tasks.py code superseded by sample_tasks.py) — if \`tasks.py\` itself is going away, the oracle may follow. But even independent of #53, the file doesn't belong at the top of the production package.

cc @payalchandak (last author, 5339211)

mmcdermott:
## Summary

Removes the pre-sampler exhaustive task-label pipeline that was superseded by `sample_tasks.py` in #41. `train.py` now consumes the sampler's long-format output directly — no collation step, no wide-to-long reshape.

Closes #53. Also closes #63.

## What got removed

- `src/every_query/tasks.py` and `tasks_config.yaml` — the entire old pipeline.
- `src/every_query/tasks_reference.py` — the regression oracle used to cross-check the new sampler against the old exhaustive-grid implementation. Per the consensus on #63 (both @gkondas and @payalchandak: "we probably don't need tasks_reference.py at all"), removed entirely now that the legacy path it mirrored is gone.
- `train.py::_collate_shard` / `collate_tasks` and the `only_preprocess` / `collate_only_held_out` knobs that only existed to serve collation.
- `EQ_generate_tasks_exhaustive` console script (plus its smoke-test entry).
- `tests/test_tasks.py`, `tests/test_collate_tasks.py`, and the oracle-dependent `TestEvaluateAgainstReference::{test_labels_match_reference, test_codes_not_in_shard_are_all_false}` (5 tests removed; covered only the deleted oracle).
- `conftest.py::task_parquet_dir` fixture (replaced by `task_labels_dir`, which emits the sampler's long-format layout).
- `scripts/sweep_layers.sh` override that passed `only_preprocess=false`.

## What got kept

- `TestEvaluateIndexDfEdgeCases` (renamed from `TestEvaluateAgainstReference`) with the two hand-crafted, oracle-independent edge-case tests: strict-`>` semantics on prediction-time vs. event-time, and missing-subject-as-censored handling.

## Config changes (`config.yaml`, `_demo_train.yaml`)

- Drop `query.task_dir`, `query.sample_times_per_subject`, `only_preprocess`, `collate_only_held_out`.
- Set `datamodule.config.task_labels_dir` directly from `$TASK_DIR` (was previously `???` and overwritten by `collate_tasks`'s return value).

## On the "training-will-be-broken" concern

Addressed in a follow-up comment below — short version: `sample_tasks.py` already emits the same long-format `(subject_id, prediction_time, boolean_value, occurs, query, duration_days)` schema that `_collate_shard` used to produce. There is no separate collation step needed; `EQ_generate_tasks` → `EQ_train` is the correct post-refactor workflow.

## Test plan

- [x] `pytest tests/ src/` — 185 passed (drop from 190 = 5 oracle-dependent tests removed with `tasks_reference.py`).
- [x] `pre-commit run --all-files` — all hooks pass.
- [ ] Reviewer check: any downstream consumer of the old `{task_dir}/collated/{hash}/...` layout needs to be migrated to the flat sampler output (see `EQ_generate_tasks` for the new layout).

## Follow-ups not in this PR

- `query.codes` still has a gitignored default (`train_codes: 10000_ID__8db2be6fadf8`). Dropping that default is tracked in #64 (Phase 5 of the refactor plan on #54).
- `upload_models.py` deletion tracked in #67 (PR #77 open).


payalchandak:
LGTM, except that train will be broken until the sample tasks is wired in @mmcdermott

mmcdermott:
@payalchandak, can you expand on how train will be broken until the sample tasks is wired in? Why is the sampling of tasks in preparation for pre-training coupled to training directly?

src/every_query/train.py
```diff
@@ -350,13 +204,6 @@ def main(cfg: DictConfig) -> float | None:
     if not isinstance(cfg.query.codes, ListConfig):
         raise ValueError("query.codes must be a list")
 
-    task_dir = collate_tasks(cfg)
```
payalchandak, line 353: sorry, all i meant was that training on new sets of tasks that have not been previously collated will not be possible. can train on tasks that have been previously collated.

mmcdermott: Great -- do we know if the updated task sampling code properly collates things? I think it does as part of its normal execution, but I want to be sure. @gkondas maybe you can check?

mmcdermott: Confirmed: `sample_tasks.py` already produces collated output — the post-refactor `EQ_generate_tasks` output is a drop-in replacement for what `_collate_shard` used to emit. No separate collation step needed between `EQ_generate_tasks` and `EQ_train`.

**Schema comparison (verified on dev @ `40b2b87`):**

Old `_collate_shard` final write (`train.py:263-277` on dev):
```
subject_id, prediction_time, boolean_value, occurs, query, duration_days
```

New `evaluate_index_df` return schema (`sample_tasks.py:271`):
```
subject_id, prediction_time, boolean_value, occurs, query, duration_days
```

Byte-for-byte identical column set + types. The dataloader that used to consume `{task_dir}/collated/{hash}/{split}/{shard}.parquet` now consumes `{sampler_out_dir}/{split}/{shard}.parquet` — same schema, flatter layout.

**Sampling-distribution caveat** (worth noting but doesn't change the answer):

- Old `_collate_shard`: for each subject, pick up to `sample_times_per_subject` prediction_times uniformly, then assign a random `(duration, code)` per sampled row.
- New `sample_tasks.run_worker`: sample `N` tasks from `(code, duration)` distribution, sample `N × M` `(subject, prediction_time)` contexts iid with replacement, zip.

Both produce long-format labeled parquets; the *distribution* over subjects differs slightly (per-subject fixed-K vs. global iid), but both are valid PT task distributions and the dataloader doesn't know or care which one produced its input.

**So the workflow after this PR is:**
```
EQ_process_data …          # tensorize MEDS
EQ_generate_tasks …        # writes long-format labels directly — no collation step
EQ_train datamodule.config.task_labels_dir=$TASK_DIR …
```

Previously `EQ_train` called `collate_tasks` inline, reading wide-per-duration parquets from `$TASK_DIR` and writing collated long parquets. That inline step is gone; the input to `EQ_train` is now already-long-format, which is exactly what `EQ_generate_tasks` emits.

gkondas: yep the task sampler removed the need for an explicity "collate" step as it outputs what collate_tasks() did before

PR #77, opened 2026-04-17
chore: remove upload_models (closes #67)

### LINKED ISSUE #67: `upload_models.py` is package-internal but only invoked via a Columbia-cluster SLURM wrapper

mmcdermott:
## Summary

Removes the Columbia-cluster GCS upload tooling that shipped inside the installable package. Per the refactor discussion on #54, cluster-specific operator tooling should live outside the package.

Closes #67.

## What got removed

- `src/every_query/upload_models.py` (198 lines) — GCS uploader; zero inbound imports from the rest of the package, only invoked via the SLURM wrapper below.
- `scripts/upload_models.sh` (42 lines) — SLURM submit script with hardcoded `gbk2114@cumc.columbia.edu` recipient and Columbia partition.
- `scripts/upload_lists/example.txt` — empty template for the above.

## What got kept

`scripts/sync.py` and `scripts/sync_config.yaml` — bidirectional GCS sync tooling that's more general (any user with a bucket can adapt the YAML). Not cluster-specific in the same way the deleted files are; kept out of #67's narrower scope. Happy to revisit if reviewers want these out too.

## Heads-up for @gkondas

This is your upload path. If you use `scripts/upload_models.sh` from the Columbia cluster, grab a copy of the three deleted files before this merges — they can live in a personal or internal-only repo (or in your `~/scripts/` dir).

## Test plan

- [x] `pytest tests/ src/` — 232 passed (no regressions)
- [x] No inbound imports to verify — grep confirms zero Python imports of `upload_models` outside the deleted shell script.


mmcdermott:
@gkondas do you want to fork a copy of the repo or otherwise copy this out into a different spot before this goes in? Not that it will be lost; it will be in the history, but wanted to flag this so you see it.

PR #86, opened 2026-04-18
fix(train): preserve config.yaml across do_overwrite re-runs (closes #31)

### LINKED ISSUE #31: `train.py`: `do_overwrite=True` deletes the config before saving the new one, losing reproducibility artifacts
## Summary

In `src/every_query/train.py`, the output-dir bookkeeping in `main()` is structured so that `config.yaml` / `resolved_config.yaml` are only written when the output dir is fresh. If the user passes `do_overwrite=True` to re-run a configuration, the old config is removed but no new config is written.

```python
output_dir = Path(cfg.output_dir)
...
os.makedirs(output_dir, exist_ok=True)

cfg_path = output_dir / "config.yaml"
ckpt_path = None
if cfg_path.exists():
    if cfg.do_overwrite:
        logger.info(f"Overwriting existing output directory {output_dir}.")
        shutil.rmtree(output_dir, ignore_errors=True)          # <-- config.yaml gone here
    elif cfg.do_resume:
        ...
    else:
        raise FileExistsError(...)
else:
    OmegaConf.save(cfg, output_dir / "config.yaml")            # <-- only runs on the "fresh dir" path
    save_resolved_config(cfg, output_dir / "resolved_config.yaml")
```

After the `do_overwrite` branch the function falls through to `torch.set_float32_matmul_precision(...)`, `instantiate(cfg.datamodule)`, and `trainer.fit(...)`. `OmegaConf.save` is never called on the overwrite path, so:

1. The output directory that \`trainer.fit\` populates no longer contains the config it was trained with.
2. `eval.py` (`src/every_query/eval.py:29`) does `train_cfg = OmegaConf.load(model_run_dir / \"resolved_config.yaml\")`. Evaluating a checkpoint that was produced by an overwrite run will fail with `FileNotFoundError`, or — worse — pick up a stale `resolved_config.yaml` from a previous run if Lightning / Hydra recreated the directory partway through.

## When this bites

- Any user re-running a hyperparameter sweep where they iterate on the same output dir with `do_overwrite=True`.
- CI / scripted reruns that always pass `do_overwrite=True` to keep the output path stable.
- Downstream `eval.py` runs against overwritten checkpoints.

## Severity

Medium. Training still succeeds, but the resulting run is no longer self-describing / reproducible, and eval against the checkpoint may break.

## Proposed fix

Restructure so that the config is written on *every* successful path that ends in `trainer.fit`, not just the pristine-dir path:

```python
output_dir = Path(cfg.output_dir)
if output_dir.is_file():
    raise NotADirectoryError(...)

cfg_path = output_dir / "config.yaml"
ckpt_path = None
if cfg_path.exists():
    if cfg.do_overwrite:
        logger.info(f"Overwriting existing output directory {output_dir}.")
    elif cfg.do_resume:
    else:
        raise FileExistsError(...)

os.makedirs(output_dir, exist_ok=True)

# Always (re)save config after we know we're going to train.
# On do_resume, skip rewriting to avoid clobbering the original (or diff-check it).
if not cfg.do_resume:
    OmegaConf.save(cfg, output_dir / "config.yaml")
    save_resolved_config(cfg, output_dir / "resolved_config.yaml")
```

## Test

A small doctest or pytest that:
1. Creates `output_dir` with a dummy `config.yaml`.
2. Calls the setup block with `do_overwrite=True`.
3. Asserts `(output_dir / "config.yaml").exists()` and its contents match the new cfg.

mmcdermott:
## Summary

`EQ_train` used to wipe `config.yaml` without re-saving it whenever `do_overwrite=True` ran against a populated output dir. Downstream tools (most visibly `eval.py`, which loads `resolved_config.yaml` from `model_run_dir`) would then fail or — worse — silently consume a stale config from a different run.

Closes #31.

## Fix

Move `os.makedirs(output_dir, exist_ok=True)` + the `OmegaConf.save` + `save_resolved_config` calls outside the `cfg_path.exists()` branch, so the config is always (re-)written after the output dir is known-good, unless `do_resume=True` (where we intentionally preserve the original run's config to keep the resume bit-identical).

## Test plan

- [x] New `TestTrainOverwriteReRun` in `tests/test_train_cli.py` runs `EQ_train` once, then again with `do_overwrite=True` into the same output_dir, asserts both `config.yaml` and `resolved_config.yaml` exist at the end.
- [x] `pytest tests/ src/` — 187 passed (was 185 + 2 new).

### DIFF
--- a/src/every_query/train.py
+++ b/src/every_query/train.py
@@ -230,12 +230,12 @@ def main(cfg: DictConfig) -> float | None:
             )
 
     # Ensure output_dir exists *after* any overwrite rmtree above, then write the config for this
-    # run.  On resume we keep the original run's config untouched so the resumed run stays
-    # bit-identical to the first.  On overwrite the previous rmtree wiped the old config; writing
-    # it here restores reproducibility for any downstream tool (e.g. ``eval.py`` loads
-    # ``resolved_config.yaml`` from the run dir).  Fixes #31.
+    # run.  On resume (without overwrite) we keep the original run's config untouched so the
+    # resumed run stays bit-identical to the first.  On overwrite the previous rmtree wiped the
+    # old config; writing it here restores reproducibility for any downstream tool (e.g.
+    # ``eval.py`` loads ``resolved_config.yaml`` from the run dir).  Fixes #31.
     os.makedirs(output_dir, exist_ok=True)
-    if not cfg.do_resume:
+    if not cfg.do_resume or cfg.do_overwrite:
         OmegaConf.save(cfg, output_dir / "config.yaml")
         save_resolved_config(cfg, output_dir / "resolved_config.yaml")
 
--- a/src/every_query/train.py
+++ b/src/every_query/train.py
@@ -213,7 +213,6 @@ def main(cfg: DictConfig) -> float | None:
     output_dir = Path(cfg.output_dir)
     if output_dir.is_file():
         raise NotADirectoryError(f"Output directory {output_dir} is a file, not a directory.")
-    os.makedirs(output_dir, exist_ok=True)
 
     cfg_path = output_dir / "config.yaml"
     ckpt_path = None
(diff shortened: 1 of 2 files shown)


mmcdermott:
@copilot review

mmcdermott:
Thanks — correct catch on the both-flags edge case, the warning at the top of main() does document that overwrite wins, and the save guard needed to match. Added TestTrainOverwriteAndResumeBothTrue in 586ec07 so the condition can't regress silently.

PR #87, opened 2026-04-18
test(eval): doctest _collect_codes with None buckets (closes #32)

### LINKED ISSUE #32: `eval.py`: `code in cfg.ood_codes` crashes with TypeError when `ood_codes` is None
## Summary

`_run_test` (`src/every_query/eval.py`) assembles the list of codes to evaluate from three optional config fields, each individually guarded against `None`:

```python
codes: list[str] = []
if cfg.id_codes is not None:
    codes += cfg.id_codes
if cfg.ood_codes is not None:
    codes += cfg.ood_codes
if cfg.manual_codes is not None:
    codes += cfg.manual_codes
```

But then, later in the same loop, the per-code bucket label unconditionally references `cfg.ood_codes`:

```python
rows.append({
    "code": code,
    "code_slug": slug,
    "bucket": "ood" if code in cfg.ood_codes else "id",
    ...
})
```

If a user runs eval with only `id_codes` / `manual_codes` set and leaves `ood_codes: null` (or omits it), every non-zero-code iteration raises `TypeError: argument of type 'NoneType' is not iterable`.

## When this bites

Any eval config that legitimately has no OOD bucket — common for in-distribution debug runs, or for per-slug checkpoints where OOD splits are evaluated by a separate command.

## Severity

Low: crash is loud (TypeError) rather than silent, but it prevents a valid configuration from running at all.

## Proposed fix

Treat a missing bucket as empty at the use site:

```python
ood_codes = cfg.ood_codes or []
...
"bucket": "ood" if code in ood_codes else "id",
```

Or normalize once at the top of the function:

```python
id_codes = cfg.id_codes or []
ood_codes = cfg.ood_codes or []
manual_codes = cfg.manual_codes or []
codes = [*id_codes, *ood_codes, *manual_codes]
```

## Test

A tiny unit test (or doctest) that constructs an `OmegaConf` with `ood_codes=None` and verifies `_run_test` labels each code with bucket `"id"` rather than raising.

mmcdermott:
## Summary

Issue #32 reported that `_run_test`'s `code in cfg.ood_codes` could crash with `TypeError: argument of type 'NoneType' is not iterable` when an eval config legitimately had no OOD bucket. An earlier refactor (the one that introduced `_collect_codes`) already fixed the use sites by switching to `cfg.X_codes or []`. This PR locks that behavior in with a doctest.

Closes #32.

## What changed

- Added a doctest block to `_collect_codes` covering: (a) all three buckets populated, (b) `ood_codes` / `manual_codes = None`, (c) duplicates across buckets (first-seen bucket label wins).

No functional change. This is a regression guard so a future refactor can't silently drop the `or []` defaults.

## Test plan

- [x] `pytest src/every_query/eval.py` — 1 passed (the new doctest).
- [x] Full suite still green.

### DIFF
--- a/src/every_query/eval.py
+++ b/src/every_query/eval.py
@@ -71,7 +71,43 @@ def _model_name(model_run_dir: str) -> str:
 
 
 def _collect_codes(cfg: DictConfig) -> tuple[list[str], dict[str, str]]:
-    """Combine id/ood/manual codes into a flat list plus a code→bucket map."""
+    """Combine id/ood/manual codes into a flat list plus a code→bucket map.
+
+    Any of ``id_codes``, ``ood_codes``, ``manual_codes`` may be ``None`` or omitted —
+    `None` is treated as an empty list (regression guard for #32, where
+    ``code in cfg.ood_codes`` crashed with ``TypeError: argument of type 'NoneType' is not
+    iterable`` for eval configs that legitimately had no OOD bucket).
+
+    Examples:
+        >>> from omegaconf import OmegaConf
+        >>> codes, bucket = _collect_codes(OmegaConf.create(
+        ...     {"id_codes": ["A", "B"], "ood_codes": ["X"], "manual_codes": ["M"]}
+        ... ))
+        >>> codes
+        ['A', 'B', 'X', 'M']
+        >>> bucket == {"A": "id", "B": "id", "X": "ood", "M": "manual"}
+        True
+
+        ``None`` for any bucket is silently treated as empty:
+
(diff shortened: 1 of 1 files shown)


mmcdermott:
@copilot review

mmcdermott:
Both points addressed in 02007fe: added an id_codes=None doctest case for bucket-guard exhaustiveness, and rewrote the prose to distinguish None (handled by or-[]) from an actually-missing key (raises ConfigAttributeError, reachable only by non-shipped configs). Thanks.

PR #88, opened 2026-04-18
fix: drop gitignored train_codes/eval_codes defaults (closes #64)

mmcdermott:
## Summary

Removes the gitignored default code-group compositions from the three shipped Hydra configs. Per [@mmcdermott on #64](https://github.com/payalchandak/EveryQuery/issues/64#issuecomment-4269211375): *"drop the default entirely is the right option."*

Closes #64.

## Before

```yaml
defaults:
  - _self_
  - train_codes: 10000_ID__8db2be6fadf8   # <-- this file is gitignored
```

`EQ_train --help` on a fresh clone:

```
hydra.errors.MissingConfigException: In 'config.yaml': Could not find 'train_codes/10000_ID__8db2be6fadf8'
```

## After

```yaml
defaults:
  - _self_

# ``train_codes`` is a required compose group.  Users must pass
# ``+train_codes=<my_codes_name>`` on the CLI with ``--config-dir`` pointing at a directory
# containing ``train_codes/<my_codes_name>.yaml`` — or inline ``query.codes=[...]`` and skip the
# compose group entirely.  See #64.
```

`EQ_train --help` works out of the box; training requires an explicit `+train_codes=…` override, which is honest about the requirement.

## Files changed

- `src/every_query/config.yaml` — drop `train_codes` default.
- `src/every_query/eval_suite/conf/eval_config.yaml` — drop `eval_codes` default.
- `src/every_query/eval_suite/conf/gen_tasks_config.yaml` — drop `eval_codes` default.
- `tests/test_cli_smoke.py` — change smoke-test overrides from `train_codes=smoke` → `+train_codes=smoke` (same for `eval_codes`), since the override no longer has a default to replace.

## Out of scope

- `sample_codes/` hardcoded MIMIC paths / non-MEDS-agnostic generator — tracked separately as #85 (Phase 4.2 of the [#54 plan](https://github.com/payalchandak/EveryQuery/issues/54#issuecomment-4271122440)).

## Test plan

- [x] `pytest tests/ src/` — 185 passed (unchanged).
- [x] `EQ_train --help` succeeds without a composed code-group (via the smoke test).
- [x] `pre-commit run --all-files` — clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #83, opened 2026-04-18
Phase 2.4: Consolidate eval pipeline into single EQ_evaluate


mmcdermott:
Parent: #54 (Phase 2.4). Depends on: #81 (`EQ_predict`) and #82 (inventory).

## Goal

Collapse the four eval CLIs — `EQ_gen_eval_index`, `EQ_gen_eval_tasks`, `EQ_evaluate`, `EQ_select_model` — into a single `EQ_evaluate` that consumes predictions written by `EQ_predict` and emits metrics. Per #82's migration plan, paper-experiments-only operations (e.g. cross-model ranking) move to `paper_experiments/`.

## Target layout

```
src/every_query/
└── evaluate/
    ├── __main__.py        → EQ_evaluate
    └── configs/
```

Previously-separate CLIs either fold in, move out, or are deleted per #82's classification:
- `EQ_gen_eval_index` — folds into `prepare_tasks/` if it survives at all (TBD by 2.3).
- `EQ_gen_eval_tasks` — expected to be obsolete once the schema-conformant tasks parquet replaces per-duration slicing.
- `EQ_evaluate` — this issue. New shape: `predictions_parquet → metrics_parquet`.
- `EQ_select_model` — likely → `paper_experiments/leaderboard/` (TBD by 2.3).

## Fixes that land along the way

- Fix #32 (`ood_codes is None` TypeError). If the ID/OOD-bucket concept survives in `evaluate/`, guard it with `cfg.ood_codes or []`; if it moves to paper_experiments, the fix moves with it.
- Fix #66 (`model_run_dirs` hardcoded list). `EQ_evaluate` post-refactor takes a single predictions-parquet path, which resolves the "why a list?" question.

## Out of scope

- `EQ_predict` itself (#81).
- Paper-experiments moves (#84).

## Acceptance criteria

- [ ] `EQ_evaluate --help` works after `pip install`, with the new shape documented.
- [ ] Four old entry points (`EQ_gen_eval_index`, `EQ_gen_eval_tasks`, `EQ_select_model`, plus the old `EQ_evaluate`) are removed from `[project.scripts]`.
- [ ] Metrics parquet schema is documented and tested.
- [ ] #32 and #66 closed (or retargeted if they moved to paper_experiments).
- [ ] Smoke test + integration test for the new entry point.

mmcdermott:
[...]

Issue #80, opened 2026-04-18
Phase 2.1 (design): Define cross-stage task-query schema via flexible_schema


mmcdermott:
Parent: #54 (Phase 2.1)

## Goal

Define the shared schema contract for rows that specify *which queries to evaluate for which subjects at which prediction times*. This is the cross-stage source of truth — produced by `EQ_prepare_tasks` and `EQ_prepare_eval` (or equivalent), consumed by `EQ_predict`.

## Deliverable

This issue is a **design doc**, not code. Outputs:
1. A proposal for the schema (column names, dtypes, allowed payloads for `task_query`).
2. An answer to "do we adopt an existing schema, extend it, or write our own?"
3. A named Python class + where it lives in the repo.

## Starting points

Per [Matthew's 2026-04-18 comment on #54](https://github.com/payalchandak/EveryQuery/issues/54#issuecomment-TBD):

- The schema should be defined using [`flexible_schema`](https://github.com/Medical-Event-Data-Standard/flexible_schema) as the base, not by subclassing `FlexibleSchema` directly.
- Instead, extend the [MEDS `LabelSchema`](https://github.com/Medical-Event-Data-Standard/meds/blob/main/src/meds/schema.py#L69), mirroring how [meds-evaluation defines its schemas](https://github.com/kamilest/meds-evaluation/blob/main/src/meds_evaluation/schema.py#L6).

So the name `FlexibleSchema` from earlier plan drafts is wrong — that's the *base class library*, not the concrete schema we're defining. Candidate names for the concrete schema, for the design doc to pick:

- `TaskQuerySchema`
- `PredictionInputSchema`
- `EQQuerySchema`
- Something else that lines up with MEDS / meds-evaluation terminology.

## Columns (starting point, to be refined)

Minimum: `subject_id`, `prediction_time`, `task_query`.

Open questions for the doc:
- Is `task_query` a flat string (a code), a struct (code + duration), or a more general expression?
- How does `duration_days` fit — a separate column, part of `task_query`, or absent?
- Does this schema also carry labels (`boolean_value`, `occurs`) for eval use, or do labels live on a separate "evaluation-time" schema that extends it?
- Split field? `split` column vs. directory layout by split?

## Out of scope

- Implementation (this is a doc). Implementation lands in a follow-up that references the decision made here.
- Predict logic (#81 issue).

## Acceptance criteria

- [ ] Design doc posted as issue comment or `docs/schema.md` in a docs-only PR.
[...]

PR #92, opened 2026-04-18
fix: mark query.codes as ??? in config.yaml (replaces #88, refs #64)

mmcdermott:
## Summary

Narrower replacement for [PR #88](https://github.com/payalchandak/EveryQuery/pull/88) after @mmcdermott's feedback there. Uses `query.codes: ???` (Hydra's mandatory-value marker) instead of the `+train_codes=…` compose-group override, and scopes down to `config.yaml` only — the eval-suite configs are intentionally left for Phase 2 of #54 to restructure.

Refs #64.

## Changes

**`src/every_query/config.yaml`:**
- Drop `defaults: - train_codes: 10000_ID__8db2be6fadf8` (gitignored default that broke `EQ_train --help` on a fresh clone).
- `query.codes: ???` — Hydra errors loudly with "missing mandatory value" if the user forgets to pass them.
- `${train_codes.codes}` interpolations under `trainer.logger.config` (wandb metadata) repointed at `${query.codes}`, which is now the single source of truth.

**`tests/test_cli_smoke.py`:**
- `EQ_train` smoke override becomes `query.codes=[]` (inline, no `+` ceremony).
- `smoke_config_dir` fixture no longer writes a `train_codes/smoke.yaml` — only `eval_codes/` remains (still needed for the two eval endpoints, which will go away in Phase 2).

## What's deliberately not here

- `eval_config.yaml` and `gen_tasks_config.yaml` still carry their `eval_codes` compose group. Converting those to `???` is awkward because ID/OOD code splits are naturally bundled into a single named YAML. Per Phase 2 of the [#54 refactor plan](https://github.com/payalchandak/EveryQuery/issues/54#issuecomment-4271122440), the eval pipeline is getting consolidated into a new `EQ_evaluate` that will redefine the input shape entirely (#83). Touching those configs in this PR would be churn.

## Why this replaces #88

[@mmcdermott on PR #88](https://github.com/payalchandak/EveryQuery/pull/88#issuecomment-TBD) flagged two points:
1. *"Any argument we're expecting users to pass should be in the config. If it is mandatory, it should have a `???` value. We don't want users to need the `+...` syntax, basically ever, if possible."* — addressed by switching to `query.codes: ???`.
2. *"Really, this looks like a bigger architectural problem, where we're maybe trying to make a small fix when this will be redundant with larger changes in the planned refactor."* — addressed by scoping down to `config.yaml` only and deferring the eval configs to Phase 2 where they're getting rewritten anyway.

## Test plan

- [x] `pytest tests/ src/` — 190 passed.
- [x] `EQ_train --help` works on a fresh clone without any override.
- [x] Running `EQ_train` without `query.codes` fails loudly with Hydra's mandatory-value error.

### DIFF
--- a/src/every_query/config.yaml
+++ b/src/every_query/config.yaml
@@ -1,6 +1,5 @@
 defaults:
   - _self_
-  - train_codes: 10000_ID__8db2be6fadf8
 
 output_dir: ${oc.env:OUTPUT_DIR}/${run_id:}/
 do_overwrite: true
@@ -8,7 +7,11 @@ do_resume: false
 seed: 140799
 
 query:
-  codes: ${train_codes.codes}
+  # Required: the list of query codes to train on.  Pass either
+  #   - inline on the CLI: EQ_train query.codes=[A,B,C]
+  #   - via a user config: EQ_train --config-dir=$MY_CFG --config-name=my_train_config
+  # (where my_train_config.yaml sets query.codes).  See #64.
+  codes: ???
 
 datamodule:
   _target_: meds_torchdata.extensions.lightning_datamodule.Datamodule
@@ -63,8 +66,8 @@ trainer:
     prefix: "" # a string to put at the beginning of metric keys
     entity: ${oc.env:WANDB_ENTITY}
     config:
-      query_codes: ${train_codes.codes}
-      num_query_codes: ${list_len:${train_codes.codes}}
+      query_codes: ${query.codes}
+      num_query_codes: ${list_len:${query.codes}}
   callbacks:
     _target_: __main__.values_as_list
     learning_rate_monitor:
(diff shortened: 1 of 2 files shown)


mmcdermott:
This is better, but I suspect is actually still worth just leaving until we do the larger refactor. Let's wait on this for a bit to consider.

mmcdermott:
Closing without merging per [your comment above](https://github.com/payalchandak/EveryQuery/pull/92#issuecomment-4273591068) — we will revisit the `query.codes` config contract as part of the larger Phase 2 refactor (#83 eval consolidation + #80 schema design) so we do not lock in a shape that is about to change.

Branch stays on the remote for now; easy to reopen if we decide the tiny fix is worth shipping before Phase 2. #64 remains open as the tracking issue.

PR #94, opened 2026-04-18
feat(train): validate config compatibility on do_resume=True (closes #91)

### LINKED ISSUE #91: train.py: do_resume=True should reject structural config drift from the resumed-from run
## What I found

`pretrain/train.py::main` (currently `src/every_query/train.py` on `dev` before #90 lands) supports `do_resume=True` but does **no** check that the on-disk config.yaml from the original run matches the config the user is now passing. The flow on resume:

1. `ckpt_path = find_checkpoint_path(output_dir)` — picks a checkpoint file.
2. `output_dir/config.yaml` is left untouched (good — preserves the original run's config).
3. `cfg` (the *new*, potentially drifted, OmegaConf) is used to instantiate the datamodule, lightning module, and trainer.
4. `trainer.fit(ckpt_path=ckpt_path)` resumes from the checkpoint.

There's no cross-check between `cfg` and `OmegaConf.load(output_dir / "config.yaml")`.

## Why this matters

- **Hard breakage** on architectural drift (`hidden_size`, `num_layers`, `num_attention_heads`, vocab size): `load_from_checkpoint` will crash at weight load. That's loud and visible.
- **Silent drift** on non-architectural knobs: `max_seq_len`, `batch_size`, `seq_sampling_strategy`, `static_inclusion_mode`, `query.codes`, `seed`. A resume with a different `max_seq_len` would produce a run that looks like the original on paper but isn't — runs posted to wandb / GCS are no longer comparable.
- **Partial drift** on training-schedule params (`max_steps`, `val_check_interval`, early-stopping patience): arguably legitimate things to override on resume. These shouldn't raise.

So the right behavior is: **fail loudly on drift in everything except an explicit deny-list of "fine to differ" keys.**

## Prior art in MEDS-EIC-AR

[`MEDS_EIC_AR/training/files.py`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/training/files.py) already solves this. Three pieces, directly portable:

1. **[`ALLOWED_DIFFERENCE_KEYS`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/training/files.py#L7-L16)** — deny-list of keys that legitimately differ on resume (the current EQ equivalent would add `trainer.max_steps` and friends):
   ```python
   ALLOWED_DIFFERENCE_KEYS = {
       "do_resume",
       "output_dir",
       "do_overwrite",
       "datamodule.num_workers",
       "trainer.logger.save_dir",
       "trainer.callbacks.model_checkpoint.dirpath",
       "trainer.default_root_dir",
       "log_dir",
   }
   ```

2. **[`diff_configs(new, old)`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/training/files.py#L63-L140)** — recursive deep-diff returning a `{dotted_key: human_message}` dict. Handles enum parameters (`seq_sampling_strategy`, `padding_side`, `static_inclusion_mode`) with case-insensitive comparison via `resolve_enum_value`, so `"RANDOM"` vs `SubsequenceSamplingStrategy.RANDOM` is not flagged as a diff.

3. **[`validate_resume_directory(output_dir, cfg)`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/training/files.py#L143-L223)** — loads on-disk `config.yaml`, runs `diff_configs`, strips anything in `ALLOWED_DIFFERENCE_KEYS`, raises `ValueError` with a formatted per-key message if anything's left.

This is strictly better than my earlier pseudocode — deep recursive diff catches drift at any nesting level, and deny-list semantics default to strict (new fields a future PR adds are automatically guarded).

## Proposed port

Copy `diff_configs` + `validate_resume_directory` into (post-#90) `src/every_query/pretrain/` as `pretrain/resume_check.py` (or similar). Tweaks for EQ:

- `ALLOWED_DIFFERENCE_KEYS` adds the training-schedule overrides we explicitly want to allow on resume:
  ```python
  ALLOWED_DIFFERENCE_KEYS = {
      "do_resume", "do_overwrite", "output_dir",
      "trainer.max_steps",            # legit to bump on resume
      "trainer.max_epochs",           # same
      "trainer.callbacks.early_stopping.patience",
      "datamodule.num_workers", "datamodule.pin_memory",
      "trainer.logger.save_dir",
      "trainer.callbacks.model_checkpoint.dirpath",
      "trainer.default_root_dir",
      "log_dir",
  }
  ```
- `STR_ENUM_PARAMS` stays identical — EQ uses the same three MTD enums.
- Call site in `train.main`:
  ```python
  elif cfg.do_resume:
      logger.info(f"Resuming training in existing output directory {output_dir}.")
      validate_resume_directory(output_dir, cfg)  # raises on structural drift
  ```

## Acceptance criteria

- [ ] `pretrain/resume_check.py` (or equivalent) landed with `diff_configs`, `validate_resume_directory`, and an `ALLOWED_DIFFERENCE_KEYS` tuned for EQ.
- [ ] Call inserted in `train.main`'s `do_resume` branch, *before* `find_checkpoint_path`.
- [ ] Regression tests:
  - resume after bumping `trainer.max_steps` — passes (it's in `ALLOWED_DIFFERENCE_KEYS`).
  - resume after changing `datamodule.config.max_seq_len` — raises with a readable diff.
  - resume with `seq_sampling_strategy: "RANDOM"` vs `SubsequenceSamplingStrategy.RANDOM` — does NOT raise (enum handling).

Found while discussing #31 / PR #86.

mmcdermott:
## Summary

Port of MEICAR's `validate_resume_directory` + `diff_configs` + `ALLOWED_DIFFERENCE_KEYS` into EQ's `train/` submodule, wired into `train.main` just before `find_checkpoint_path` in the `do_resume` branch.

Closes #91.

## What lands

- **`src/every_query/train/resume_check.py`** — the ported helpers. [MEICAR reference](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/training/files.py) is cited in the module docstring.
- **Wire-in** in `train/train.py::main` — one call to `validate_resume_directory(output_dir, cfg)` at the top of the `do_resume` branch, before `find_checkpoint_path` so we fail fast (no checkpoint load on a drifted config).
- **Regression test** `TestTrainResumeRejectsStructuralDrift::test_max_seq_len_drift_is_rejected` — initial training, then resume with `datamodule.config.max_seq_len=32` (vs. 64 in `_demo_train.yaml`), asserts the subprocess fails with `max_seq_len` named in the error.
- The existing `TestTrainResume` already exercises the complementary positive case (resume with a bumped `trainer.max_steps` — which is in `ALLOWED_DIFFERENCE_KEYS` — succeeds).

## EQ-side tweak to ALLOWED_DIFFERENCE_KEYS

MEICAR's list is purely paths + lifecycle flags + `num_workers`. Extended here with training-schedule knobs that are legitimate things to override on resume:

```python
"trainer.max_steps",
"trainer.max_epochs",
"trainer.callbacks.early_stopping.patience",
"datamodule.pin_memory",
```

Common workflow: a run hits `max_steps`, you realize you want to train longer, so you bump `trainer.max_steps` on resume. That shouldn't be a structural-drift rejection.

## Enum handling (inherited from MEICAR, unchanged)

`seq_sampling_strategy`, `padding_side`, `static_inclusion_mode` are compared via `resolve_enum_value` so `"RANDOM"` vs. `SubsequenceSamplingStrategy.RANDOM` vs. `"random"` don't false-positive. Covered by the ported doctest.

## Test plan

- [x] `pytest tests/ src/` — 194 passed (191 baseline + 2 new doctests in `resume_check.py` + 1 new regression test).
- [x] Positive case (resume with bumped `max_steps`) passes via the existing `TestTrainResume` fixture.

### DIFF
--- a/src/every_query/train/train.py
+++ b/src/every_query/train/train.py
@@ -222,6 +222,9 @@ def main(cfg: DictConfig) -> float | None:
             shutil.rmtree(output_dir, ignore_errors=True)
         elif cfg.do_resume:
             logger.info(f"Resuming training in existing output directory {output_dir}.")
+            from every_query.train.resume_check import validate_resume_directory
+
+            validate_resume_directory(output_dir, cfg)
             ckpt_path = find_checkpoint_path(output_dir)
         else:
             raise FileExistsError(
--- a/src/every_query/train/train.py
+++ b/src/every_query/train/train.py
@@ -14,6 +14,8 @@ from lightning.pytorch import seed_everything
 from MEDS_transforms.configs.utils import OmegaConfResolver
 from omegaconf import DictConfig, ListConfig, OmegaConf
 
+from every_query.train.resume_check import validate_resume_directory
+
 logger = logging.getLogger(__name__)
 
 _RUN_ID = None
@@ -222,8 +224,6 @@ def main(cfg: DictConfig) -> float | None:
             shutil.rmtree(output_dir, ignore_errors=True)
         elif cfg.do_resume:
             logger.info(f"Resuming training in existing output directory {output_dir}.")
-            from every_query.train.resume_check import validate_resume_directory
-
             validate_resume_directory(output_dir, cfg)
             ckpt_path = find_checkpoint_path(output_dir)
         else:
--- a/src/every_query/train/resume_check.py
+++ b/src/every_query/train/resume_check.py
@@ -1,7 +1,7 @@
 """Validate that a ``do_resume=True`` invocation matches the resumed-from run's config.
 
 Ported from [MEDS_EIC_AR's ``training/files.py``](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/training/files.py)
-with one substantive EQ-side tweak: the ``ALLOWED_DIFFERENCE_KEYS`` deny-list adds a handful
+with one substantive EQ-side tweak: the ``ALLOWED_DIFFERENCE_KEYS`` allow-list adds a handful
 of training-schedule knobs (``trainer.max_steps``, ``trainer.max_epochs``,
 ``trainer.callbacks.early_stopping.patience``, ``datamodule.pin_memory``) that are *legitimate*
 things to override on resume — bumping step count to train longer than originally planned is a
--- a/src/every_query/train/resume_check.py
+++ b/src/every_query/train/resume_check.py
@@ -1,0 +1,196 @@
+"""Validate that a ``do_resume=True`` invocation matches the resumed-from run's config.
+
+Ported from [MEDS_EIC_AR's ``training/files.py``](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/training/files.py)
+with one substantive EQ-side tweak: the ``ALLOWED_DIFFERENCE_KEYS`` deny-list adds a handful
+of training-schedule knobs (``trainer.max_steps``, ``trainer.max_epochs``,
+``trainer.callbacks.early_stopping.patience``, ``datamodule.pin_memory``) that are *legitimate*
+things to override on resume — bumping step count to train longer than originally planned is a
+common workflow.
+
+See #91 for the motivation and #31 for the original ``do_overwrite`` config-preservation bug
+this complements.
+"""
+
+from pathlib import Path
+from typing import Any
+
+from meds_torchdata.types import PaddingSide, StaticInclusionMode, SubsequenceSamplingStrategy
+from omegaconf import DictConfig, OmegaConf
+
+ALLOWED_DIFFERENCE_KEYS = {
+    # Lifecycle flags — by definition different on resume.
+    "do_resume",
+    "do_overwrite",
+    # Paths that legitimately differ between runs of the same config.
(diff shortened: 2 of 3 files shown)


mmcdermott:
@copilot review

mmcdermott:
@copilot review

src/every_query/train/train.py
```diff
@@ -222,6 +222,9 @@ def main(cfg: DictConfig) -> float | None:
             shutil.rmtree(output_dir, ignore_errors=True)
         elif cfg.do_resume:
             logger.info(f"Resuming training in existing output directory {output_dir}.")
+            from every_query.train.resume_check import validate_resume_directory
```
mmcdermott: do this sort of module import at the top. I know it is only necessary sometimes, but it should be basically no cost, right?

mmcdermott: Agreed — hoisted to module-level in 4cafd63.  No circular-import risk with `resume_check` and the cost is trivial.

PR #96, opened 2026-04-18
feat(data): TaskQuerySchema — cross-stage contract for task-query rows (closes #80)

### LINKED ISSUE #80: Phase 2.1 (design): Define cross-stage task-query schema via flexible_schema

### LINKED ISSUE #122: Schema refactor: collapse EQ sampler's two label columns into one nullable `boolean_value`
## ⚠️ Scope correction (2026-04-19)

Original version of this issue proposed dropping the censor prediction head and renaming `query` to `code`.  **Both were wrong** per clarification on #96.  Rewritten below.

## Motivation

The sampler's output currently carries two label columns:

- `boolean_value` — overloaded to mean *censored* (subject data ended before we could observe the outcome).
- `occurs` — the real "did the event fire" label.

Readers of MEDS-shaped parquets reasonably expect `boolean_value` to mean *the label value*, not *a censoring flag*.  I hit this the hard way writing the training-validity test in #119 — there's a big inline comment in `tests/test_training_validity.py` specifically documenting the gotcha.

But the sampler's two columns aren't independent: `occurs` is only meaningful **conditional on** `boolean_value` (the censoring flag) being `False`.  When censored, the `occurs` value is don't-care.  That structural conditioning is exactly what a **nullable** `boolean_value` expresses idiomatically in MEDS's `LabelSchema`:

| `boolean_value` | Meaning |
|---|---|
| `null` | Censored — we can't answer the query for this patient |
| `True` | Not censored AND event occurred in the window |
| `False` | Not censored AND event did not occur |

One column, three states, semantically aligned with MEDS conventions.

### Model code: **unchanged**

This is the key correction vs. the original issue.  The existing loss wiring in `model/model.py:614-621`

```python
censor_loss = self._get_loss(censor_logits, batch.censor, mask=None)
occurs_loss = self._get_loss(occurs_logits, batch.occurs, mask=~batch.censor)
```

keeps working as-is.  Both prediction heads stay — `censor` predicts "can we answer?" and `occurs` predicts "conditioned on being able to answer, what would the answer be?"  Both are meaningful learning targets.  Only the **source** of `batch.censor` / `batch.occurs` changes (one column → two derived tensors instead of two columns → two directly-copied tensors).

**Checkpoints stay loadable.**  `state_dict` keys are unchanged.  No backward-compat break for existing trained `.ckpt` files.

### Sampler (`evaluate_index_df`): single-column output

Current:
```python
.with_columns(
    pl.col("_censored").alias("boolean_value"),
    (pl.col("_censored").not_() & event_in_window).alias("occurs"),
)
.select("subject_id", "prediction_time", "boolean_value", "occurs", "query", "duration_days")
```

Target:
```python
.with_columns(
    pl.when(pl.col("_censored"))
    .then(None)
    .otherwise(event_in_window)
    .alias("boolean_value"),
)
.select("subject_id", "prediction_time", "boolean_value", "query", "duration_days")
```

No more `occurs` column in the emitted parquet.

### Dataset (`data/dataset.py::collate`): derive two tensors from one column

Current (paraphrased):
```python
out["censor"] = out[self.LABEL_COL]  # LABEL_COL == "boolean_value"
out["occurs"] = torch.Tensor([item["occurs"] for item in batch]).long()
```

Target:
```python
is_censored = out["boolean_value"].isnan()   # or equivalent null check for the dtype
out["censor"] = is_censored
out["occurs"] = out["boolean_value"].fill_null(False).long()  # placeholder where censored
```

Model's loss mask (`~batch.censor`) still zeros out the placeholder rows correctly.

### Not renaming `query` → `code`

Keeping `query` as the sampler's column name.  The future plan is to generalise `query` into a struct (ANY/ALL compound queries, event-bounded durations) — using `code` now would imply a flat-string contract that doesn't hold long-term.

**`TaskQuerySchema` in #96 needs its column renamed from `code` → `query`** to match.  That's a one-line change in `src/every_query/data/schema.py`.

## Affected call sites

| Component | Change |
|---|---|
| `data/schema.py::TaskQuerySchema` | `code` field renamed to `query` |
| `generate_tasks/sample_tasks.py::evaluate_index_df` | Emit nullable `boolean_value`; drop `occurs` column |
| `data/dataset.py::EveryQueryPytorchDataset.collate` | Derive `batch.censor` + `batch.occurs` from `boolean_value` |
| `data/dataset.py::LABEL_TENSOR_NAMES` | Drop `occurs` from the tuple |
| `model/model.py` | **Unchanged** |
| `tests/test_sample_tasks.py::TestEvaluateIndexDfEdgeCases` | Update expected `(boolean_value=None\|True\|False)` semantics |
| `tests/test_generate_tasks.py::test_labels_match_ground_truth` | Recompute single `boolean_value` per the new semantics |
| `tests/test_training_validity.py` | Synthetic labels: `boolean_value=<occurs_truth>` directly (no subjects are censored in this test); drop the inline gotcha comment |

## Risks

1. **Loss-mask regression** is still the biggest one — if `batch.censor` is derived incorrectly from `boolean_value`, `mask=~batch.censor` ends up wrong and training silently diverges.  #119 is the strongest existing guard (HOT-COLD AUROC drops to chance if the mask inverts or goes uniformly True).
2. **Paper-experiment scripts** may still reference the `occurs` column directly.  Audit `paper_experiments/` before the sampler change.  (Post-grep confirms nothing in `paper_experiments/` touches `occurs` today.)
3. **Checkpoint backward-compat**: preserved by keeping both heads.  No action needed.

## Landing plan

Bundle the migration into #96 itself (per owner preference — see #96 review thread).  Sequencing:

1. Merge **#119** first (on current semantics) so dev has the training-validity safety net before the migration commit.
2. Expand **#96** to include the full migration above.
3. Rebase #112, #113, #119's small test-only delta onto the expanded #96.
4. Merge expanded #96.

cc @mmcdermott @payalchandak

mmcdermott:
## Summary

Closes #80, closes #122.  Defines the cross-stage task-query schema AND wires it through the sampler + dataset so it's actually load-bearing — not just an unused type definition.

## What lands

- **`src/every_query/data/schema.py`** — `TaskQuerySchema(LabelSchema)` with two required columns plus a nullability override on the inherited `boolean_value`:
  - `query: Required(pa.large_string(), nullable=False)` — the MEDS code the query asks about.  `large_string` matches what polars' `Utf8` serialises to when converted to arrow (verified via `df.to_arrow().schema`), so producers emit schema-conformant output natively and `validate()` / `align()` work without type coercion.
  - `duration_days: Required(pa.float32(), nullable=False)` — fractional-days-capable horizon.
  - `boolean_value: Optional(pa.bool_(), nullable=True)` — overrides MEDS `LabelSchema`'s `nullable=NONE` to allow nulls, since the EveryQuery convention collapses censor + occurs into this single three-valued column (closes #122): `null` = censored, `True` = event occurred, `False` = no event and not censored.
- **Sampler wires the schema through** (`generate_tasks/sample_tasks.py`):
  - `build_index_df` emits `Float32` `duration_days` natively (was `Int64`).
  - `evaluate_index_df` emits the collapsed nullable `boolean_value` (null/True/False) instead of the old `(boolean_value, occurs)` pair.
  - Empty-input fast-path delegates to the new `empty_task_query_df()` helper in `every_query.data.schema` (keyed off the schema's exported `_name` attrs) — no hand-rolled polars-schema dict.
  - `run_worker` calls `TaskQuerySchema.align(labeled.to_arrow())` at the write boundary.  `align()` over `validate()` because it reorders columns + coerces polars-native types (`Utf8` → `large_string`) that producers emit naturally, while still erroring on missing/extra columns or unreconcilable type drift — matching the pattern MEDS-transforms / meds-testing-helpers use internally.
  - All hardcoded `"subject_id"` / `"prediction_time"` / `"query"` / `"duration_days"` / `"boolean_value"` string literals replaced with the schema's exported `{col}_name` class attributes (same for the events-frame columns, which pull from `meds.DataSchema.{col}_name`).
- **Dataset reads the collapsed column + splits it back at read time** (`data/dataset.py::EveryQueryPytorchDataset.__init__`): derives `schema_df["boolean_value"] = is_null()` (censor indicator) + `schema_df["occurs"] = fill_null(False)`.  Batch-level API unchanged — `EveryQueryBatch.censor` + `EveryQueryBatch.occurs` are the names downstream code uses.
- **Other producers** (`predict/external_tasks/aces_to_eq.py`, `evaluate/gen_task.py`) emit the collapsed column too.  `aces_to_eq` now also takes `duration_days` as a required config knob and stamps it into the output, then aligns via `TaskQuerySchema.align(...)` at write.
- **Dataset back-compat branch** — `EveryQueryPytorchDataset.__init__` detects legacy two-column parquets via the presence of an `occurs` column and passes them through untouched; new collapsed parquets (nullable `boolean_value` only) are expanded to the `(boolean_value, occurs)` pair the super class consumes.
- **Tests use `TaskQuerySchema.validate()` for schema assertions** — in `tests/test_sample_tasks.py` and `tests/test_eval_suite.py` — replacing hand-rolled column-set checks.  Test fixtures (`conftest.py::task_labels_dir`, `demo_dataset`, `tests/training_validity/`) updated to emit the collapsed column with correct `Float32` dtype for `duration_days`.
- **Training-validity README doctest** updated to show `boolean_value`-only tables with `null` entries for censored rows; "Gotchas" section rewritten.

## Design choices locked in

- **Extends `LabelSchema`, not `PyArrowSchema` directly.** Matches the pattern `meds-evaluation`'s `PredictionSchema` uses.  Inherited columns: `subject_id`, `prediction_time`, `boolean_value` (now `nullable=True`), `integer_value`, `float_value`, `categorical_value`.
- **Column named `query`, not `code`.** Matches what the sampler / dataset / `EveryQueryBatch.query` already call this column; `code` would collide with `EveryQueryBatch.code` (the inherited event-sequence-token tensor — a distinct thing semantically).
- **`boolean_value` collapsed to a single nullable column** rather than two columns (`boolean_value` + `occurs`).  Cleaner schema contract, and the dataset can always derive the split back if downstream code needs it.  Mirrors MEDS's own label conventions.
- **No `task_query` struct yet.**  Keeping the query flat (just `query` + `duration_days`) honors the narrow-initial-scope direction from #54.  Compound ANY/ALL grammars, structured-query payloads, etc. are extension points for follow-up PRs.

## Test plan

- [x] `uv run pytest` — 206 passed, 1 deselected (slow) in 136s locally.
- [x] CI green (code-quality + tests 3.11/3.12 all pass on 4bc0905).

## Follow-ups

- `EQ_predict` (draft PR [#99](https://github.com/payalchandak/EveryQuery/pull/99)) now inherits `TaskQuerySchema` for `PredictionSchema` + emits `censor_prob` + `occurs_prob` (two-head probabilities).  Unblocked by this PR.
- `EQ_evaluate` consolidation (draft PR [#100](https://github.com/payalchandak/EveryQuery/pull/100)) consumes `PredictionSchema` parquets.  Unblocked by this PR + #99.
- [#129](https://github.com/payalchandak/EveryQuery/issues/129) tracks renaming `occurs_prob → label_prob` once non-occurrence task types land (post-NeurIPS).

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -317,7 +317,8 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
                     "prediction_time": _PRED_TIMES[subj],
                     "boolean_value": bv,
                     "query": query_code,
-                    "duration_days": 30,
+                    # Float to match ``TaskQuerySchema.duration_days`` (``pa.float32``).
+                    "duration_days": 30.0,
                 }
             )
 
@@ -327,7 +328,7 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
                 "subject_id": pl.Int64,
                 "boolean_value": pl.Boolean,
                 "query": pl.Utf8,
-                "duration_days": pl.Int64,
+                "duration_days": pl.Float32,
             }
         )
         df.write_parquet(split_dir / "0.parquet")
@@ -369,7 +370,7 @@ def demo_dataset(
             "subject_id": pl.Int64,
             "boolean_value": pl.Boolean,
             "query": pl.Utf8,
-            "duration_days": pl.Float64,
+            "duration_days": pl.Float32,  # matches TaskQuerySchema
         }
     )
     df.write_parquet(collated_dir / "0.parquet")
--- a/conftest.py
+++ b/conftest.py
@@ -289,7 +289,7 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
 
         {task_labels_dir}/{split}/{shard}.parquet
 
-    Columns: ``subject_id, prediction_time, boolean_value, occurs, code, duration_days``.
+    Columns: ``subject_id, prediction_time, boolean_value, occurs, query, duration_days``.
     """
     task_dir = tmp_path_factory.mktemp("eq_task_labels")
 
@@ -305,7 +305,7 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
                     "prediction_time": _PRED_TIMES[subj],
                     "boolean_value": i % 2 == 0,
                     "occurs": i % 3 != 0,
-                    "code": query_code,
+                    "query": query_code,
                     "duration_days": 30,
                 }
             )
@@ -316,7 +316,7 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
                 "subject_id": pl.Int64,
                 "boolean_value": pl.Boolean,
                 "occurs": pl.Boolean,
-                "code": pl.Utf8,
+                "query": pl.Utf8,
                 "duration_days": pl.Int64,
             }
         )
@@ -342,7 +342,7 @@ def demo_dataset(
                 "prediction_time": _PRED_TIMES[subj],
                 "boolean_value": i % 2 == 0,
                 "occurs": i % 3 != 0,
-                "code": query_code,
+                "query": query_code,
                 "duration_days": 30.0,
             }
         )
@@ -353,7 +353,7 @@ def demo_dataset(
             "subject_id": pl.Int64,
             "boolean_value": pl.Boolean,
             "occurs": pl.Boolean,
-            "code": pl.Utf8,
+            "query": pl.Utf8,
             "duration_days": pl.Float64,
         }
     )
--- a/conftest.py
+++ b/conftest.py
@@ -289,7 +289,7 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
 
         {task_labels_dir}/{split}/{shard}.parquet
 
-    Columns: ``subject_id, prediction_time, boolean_value, occurs, query, duration_days``.
+    Columns: ``subject_id, prediction_time, boolean_value, occurs, code, duration_days``.
     """
     task_dir = tmp_path_factory.mktemp("eq_task_labels")
 
@@ -305,7 +305,7 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
                     "prediction_time": _PRED_TIMES[subj],
                     "boolean_value": i % 2 == 0,
                     "occurs": i % 3 != 0,
-                    "query": query_code,
+                    "code": query_code,
                     "duration_days": 30,
                 }
             )
@@ -316,7 +316,7 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
                 "subject_id": pl.Int64,
                 "boolean_value": pl.Boolean,
                 "occurs": pl.Boolean,
-                "query": pl.Utf8,
+                "code": pl.Utf8,
                 "duration_days": pl.Int64,
             }
         )
@@ -342,7 +342,7 @@ def demo_dataset(
                 "prediction_time": _PRED_TIMES[subj],
                 "boolean_value": i % 2 == 0,
                 "occurs": i % 3 != 0,
-                "query": query_code,
+                "code": query_code,
                 "duration_days": 30.0,
             }
         )
@@ -353,7 +353,7 @@ def demo_dataset(
             "subject_id": pl.Int64,
             "boolean_value": pl.Boolean,
             "occurs": pl.Boolean,
-            "query": pl.Utf8,
+            "code": pl.Utf8,
             "duration_days": pl.Float64,
         }
     )
--- a/conftest.py
+++ b/conftest.py
@@ -289,7 +289,9 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
 
         {task_labels_dir}/{split}/{shard}.parquet
 
-    Columns: ``subject_id, prediction_time, boolean_value, occurs, query, duration_days``.
+    Columns: ``subject_id, prediction_time, boolean_value, query, duration_days``.
+    Conforms to ``TaskQuerySchema`` — ``boolean_value`` is nullable (null=censored,
+    True=event occurred, False=no event and not censored).
     """
     task_dir = tmp_path_factory.mktemp("eq_task_labels")
 
@@ -299,12 +301,21 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
 
         rows = []
         for i, (subj, query_code) in enumerate((s, q) for s in subjects for q in _QUERY_CODES):
+            # Alternate the three-valued state so downstream tests see each:
+            #   i % 3 == 0 → censored  (null)
+            #   i % 3 == 1 → occurred  (True)
+            #   i % 3 == 2 → no event  (False)
+            if i % 3 == 0:
+                bv = None
+            elif i % 3 == 1:
+                bv = True
+            else:
+                bv = False
             rows.append(
                 {
                     "subject_id": subj,
                     "prediction_time": _PRED_TIMES[subj],
-                    "boolean_value": i % 2 == 0,
-                    "occurs": i % 3 != 0,
+                    "boolean_value": bv,
                     "query": query_code,
                     "duration_days": 30,
                 }
@@ -315,7 +326,6 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
                 "prediction_time": pl.Datetime("us"),
                 "subject_id": pl.Int64,
                 "boolean_value": pl.Boolean,
-                "occurs": pl.Boolean,
                 "query": pl.Utf8,
                 "duration_days": pl.Int64,
             }
@@ -336,12 +346,18 @@ def demo_dataset(
 
     rows = []
     for i, (subj, query_code) in enumerate((s, q) for s in _TRAIN_SUBJECTS for q in _QUERY_CODES):
+        # Three-valued nullable boolean_value per TaskQuerySchema.
+        if i % 3 == 0:
+            bv = None
+        elif i % 3 == 1:
+            bv = True
+        else:
+            bv = False
         rows.append(
             {
                 "subject_id": subj,
                 "prediction_time": _PRED_TIMES[subj],
-                "boolean_value": i % 2 == 0,
-                "occurs": i % 3 != 0,
+                "boolean_value": bv,
                 "query": query_code,
                 "duration_days": 30.0,
             }
@@ -352,7 +368,6 @@ def demo_dataset(
             "prediction_time": pl.Datetime("us"),
             "subject_id": pl.Int64,
             "boolean_value": pl.Boolean,
-            "occurs": pl.Boolean,
             "query": pl.Utf8,
             "duration_days": pl.Float64,
         }
--- a/src/every_query/data/dataset.py
+++ b/src/every_query/data/dataset.py
@@ -305,9 +305,9 @@ class EveryQueryPytorchDataset(MEDSPytorchDataset):
         if "occurs" in label_names:
             group_cols.append("occurs")
             out_cols.append("occurs")
-        if "query" in label_names:
-            group_cols.append("query")
-            out_cols.append("query")
+        if "code" in label_names:
+            group_cols.append("code")
+            out_cols.append("code")
         if "duration_days" in label_names:
             group_cols.append("duration_days")
             out_cols.append("duration_days")
@@ -332,12 +332,17 @@ class EveryQueryPytorchDataset(MEDSPytorchDataset):
                 .alias(LabelSchema.prediction_time_name)
             )
 
-        # Extra task annotations
+        # Extra task annotations.  Parquet column name is ``code`` (matches
+        # ``TaskQuerySchema`` in ``every_query.data.schema``); the Python attribute is kept
+        # as ``self.query`` / ``self.has_query`` because the batch-level counterpart
+        # ``EveryQueryBatch.query`` already occupies that name — and the inherited
+        # ``EveryQueryBatch.code`` is the event-sequence-level code tensor, not the query
+        # code.  Renaming either would collide.
         self.has_occurs: bool = "occurs" in self.schema_df.collect_schema().names()
-        self.has_query: bool = "query" in self.schema_df.collect_schema().names()
+        self.has_query: bool = "code" in self.schema_df.collect_schema().names()
         self.has_duration_days: bool = "duration_days" in self.schema_df.collect_schema().names()
         self.occurs = self.schema_df["occurs"] if self.has_occurs else None
-        self.query = self.schema_df["query"] if self.has_query else None
+        self.query = self.schema_df["code"] if self.has_query else None
         self.duration_days = self.schema_df["duration_days"] if self.has_duration_days else None
         # Load code vocabulary mapping (string code -> integer vocab index) for encoding queries
         try:
@@ -364,7 +369,7 @@ class EveryQueryPytorchDataset(MEDSPytorchDataset):
         def read_df(fp: Path) -> pl.DataFrame:
             schema = pq.read_schema(fp)
             extras = []
-            for extra_col in [self.LABEL_COL, "occurs", "query", "duration_days"]:
+            for extra_col in [self.LABEL_COL, "occurs", "code", "duration_days"]:
                 if extra_col in schema.names:
                     extras.append(extra_col)
             label_cols = [*required_cols, *extras]
--- a/src/every_query/data/dataset.py
+++ b/src/every_query/data/dataset.py
@@ -332,10 +332,41 @@ class EveryQueryPytorchDataset(MEDSPytorchDataset):
                 .alias(LabelSchema.prediction_time_name)
             )
 
+        # Split the collapsed nullable ``boolean_value`` label into the separate
+        # ``(censor, occurs)`` pair the model's loss code consumes.
+        #
+        # On-disk schema (TaskQuerySchema in every_query.data.schema):
+        #     boolean_value: nullable bool
+        #         null  → censored (observation ended before we could see the outcome)
+        #         True  → event occurred in the duration window
+        #         False → no event in the window, and not censored
+        #
+        # Batch-level API (EveryQueryBatch):
+        #     censor: bool          ← boolean_value.is_null()
+        #     occurs: bool/long     ← boolean_value.fill_null(False) (masked on censored rows)
+        #
+        # Done here (after super().__init__() has already read the label parquet into
+        # self.schema_df) so the downstream getitem / collate path sees non-null tensors
+        # — torch's bool→tensor conversion doesn't tolerate nulls.
+        schema_cols = self.schema_df.collect_schema().names()
+        if "boolean_value" in schema_cols:
+            raw = pl.col("boolean_value")
+            self.schema_df = self.schema_df.with_columns(
+                raw.is_null().alias("boolean_value"),
+                raw.fill_null(False).alias("occurs"),
+            )
+            schema_cols = self.schema_df.collect_schema().names()
+            # Super cached the pre-collapse ``boolean_value`` Series into ``self.labels``
+            # at its ``__init__`` time; re-point it at the post-processed (non-null,
+            # censor-indicator) column so ``_seeded_getitem`` and ``collate`` see the
+            # correct values.
+            if getattr(self, "has_task_labels", False):
+                self.labels = self.schema_df["boolean_value"]
+
         # Extra task annotations
-        self.has_occurs: bool = "occurs" in self.schema_df.collect_schema().names()
-        self.has_query: bool = "query" in self.schema_df.collect_schema().names()
-        self.has_duration_days: bool = "duration_days" in self.schema_df.collect_schema().names()
+        self.has_occurs: bool = "occurs" in schema_cols
+        self.has_query: bool = "query" in schema_cols
+        self.has_duration_days: bool = "duration_days" in schema_cols
         self.occurs = self.schema_df["occurs"] if self.has_occurs else None
         self.query = self.schema_df["query"] if self.has_query else None
         self.duration_days = self.schema_df["duration_days"] if self.has_duration_days else None
@@ -364,6 +395,10 @@ class EveryQueryPytorchDataset(MEDSPytorchDataset):
         def read_df(fp: Path) -> pl.DataFrame:
             schema = pq.read_schema(fp)
             extras = []
+            # "occurs" is not an on-disk column after the nullable-label collapse; it's
+            # derived in __init__ from boolean_value.is_null().  Keep it in the list for
+            # backward compatibility with any legacy label parquets that still carry it
+            # explicitly — harmless if absent.
             for extra_col in [self.LABEL_COL, "occurs", "query", "duration_days"]:
                 if extra_col in schema.names:
                     extras.append(extra_col)
--- a/src/every_query/data/dataset.py
+++ b/src/every_query/data/dataset.py
@@ -305,9 +305,9 @@ class EveryQueryPytorchDataset(MEDSPytorchDataset):
         if "occurs" in label_names:
             group_cols.append("occurs")
             out_cols.append("occurs")
-        if "code" in label_names:
-            group_cols.append("code")
-            out_cols.append("code")
+        if "query" in label_names:
+            group_cols.append("query")
+            out_cols.append("query")
         if "duration_days" in label_names:
             group_cols.append("duration_days")
             out_cols.append("duration_days")
@@ -332,17 +332,12 @@ class EveryQueryPytorchDataset(MEDSPytorchDataset):
                 .alias(LabelSchema.prediction_time_name)
             )
 
-        # Extra task annotations.  Parquet column name is ``code`` (matches
-        # ``TaskQuerySchema`` in ``every_query.data.schema``); the Python attribute is kept
-        # as ``self.query`` / ``self.has_query`` because the batch-level counterpart
-        # ``EveryQueryBatch.query`` already occupies that name — and the inherited
-        # ``EveryQueryBatch.code`` is the event-sequence-level code tensor, not the query
-        # code.  Renaming either would collide.
+        # Extra task annotations
         self.has_occurs: bool = "occurs" in self.schema_df.collect_schema().names()
-        self.has_query: bool = "code" in self.schema_df.collect_schema().names()
+        self.has_query: bool = "query" in self.schema_df.collect_schema().names()
         self.has_duration_days: bool = "duration_days" in self.schema_df.collect_schema().names()
         self.occurs = self.schema_df["occurs"] if self.has_occurs else None
-        self.query = self.schema_df["code"] if self.has_query else None
+        self.query = self.schema_df["query"] if self.has_query else None
         self.duration_days = self.schema_df["duration_days"] if self.has_duration_days else None
         # Load code vocabulary mapping (string code -> integer vocab index) for encoding queries
         try:
@@ -369,7 +364,7 @@ class EveryQueryPytorchDataset(MEDSPytorchDataset):
         def read_df(fp: Path) -> pl.DataFrame:
             schema = pq.read_schema(fp)
             extras = []
-            for extra_col in [self.LABEL_COL, "occurs", "code", "duration_days"]:
+            for extra_col in [self.LABEL_COL, "occurs", "query", "duration_days"]:
                 if extra_col in schema.names:
                     extras.append(extra_col)
             label_cols = [*required_cols, *extras]
--- a/src/every_query/data/dataset.py
+++ b/src/every_query/data/dataset.py
@@ -332,24 +332,32 @@ class EveryQueryPytorchDataset(MEDSPytorchDataset):
                 .alias(LabelSchema.prediction_time_name)
             )
 
-        # Split the collapsed nullable ``boolean_value`` label into the separate
-        # ``(censor, occurs)`` pair the model's loss code consumes.
+        # Derive the ``(censor, occurs)`` pair the model's loss code consumes, from
+        # whichever on-disk label shape the parquet has.  Two shapes are supported:
         #
-        # On-disk schema (TaskQuerySchema in every_query.data.schema):
-        #     boolean_value: nullable bool
-        #         null  → censored (observation ended before we could see the outcome)
-        #         True  → event occurred in the duration window
-        #         False → no event in the window, and not censored
+        # 1. **Collapsed (new, ``TaskQuerySchema``-shaped)** — the parquet has a single
+        #    nullable ``boolean_value`` column:
+        #        null  → censored (observation ended before we could see the outcome)
+        #        True  → event occurred in the duration window
+        #        False → no event in the window, and not censored
+        #    We split it into ``boolean_value = is_null()`` (censor indicator) and
+        #    ``occurs = fill_null(False)``.
         #
-        # Batch-level API (EveryQueryBatch):
-        #     censor: bool          ← boolean_value.is_null()
-        #     occurs: bool/long     ← boolean_value.fill_null(False) (masked on censored rows)
+        # 2. **Legacy two-column** — the parquet has both ``boolean_value`` (meaning
+        #    "censored") AND an explicit ``occurs`` column.  We leave both alone;
+        #    super's ``self.labels`` cache is already the censor indicator.
         #
-        # Done here (after super().__init__() has already read the label parquet into
-        # self.schema_df) so the downstream getitem / collate path sees non-null tensors
-        # — torch's bool→tensor conversion doesn't tolerate nulls.
+        # Detection: presence of an explicit ``occurs`` column marks the legacy shape
+        # — the collapsed shape never emits it (see ``evaluate_index_df``).  This
+        # preserves backward compat with any label parquets produced by pre-#96
+        # pipelines still on disk.
+        #
+        # Done after super().__init__() has read the label parquet into self.schema_df
+        # so the downstream getitem / collate path sees non-null tensors (torch's
+        # bool→tensor conversion doesn't tolerate nulls).
         schema_cols = self.schema_df.collect_schema().names()
-        if "boolean_value" in schema_cols:
+        is_legacy_two_col = "occurs" in schema_cols
+        if "boolean_value" in schema_cols and not is_legacy_two_col:
             raw = pl.col("boolean_value")
             self.schema_df = self.schema_df.with_columns(
                 raw.is_null().alias("boolean_value"),
--- a/src/every_query/data/schema.py
+++ b/src/every_query/data/schema.py
@@ -24,7 +24,7 @@ evolve.
 """
 
 import pyarrow as pa
-from flexible_schema import Required
+from flexible_schema import Optional, Required
 from meds import LabelSchema
 
 
@@ -38,9 +38,15 @@ class TaskQuerySchema(LabelSchema):
     both inference input (no label) and evaluation input (label filled in) without a branch.
 
     Attributes:
-        query: The MEDS code the query asks about.  Named ``query`` rather than ``code`` to
-            match the column name already used throughout the sampler / dataset / batch
-            layer (``EveryQueryBatch.code`` is the distinct event-sequence-token tensor).
+        query: The MEDS code the query asks about.  Stored as ``pa.large_string``
+            (polars' ``Utf8`` serializes to ``large_string`` when a DataFrame is
+            converted to arrow, so this matches producer output natively and
+            ``TaskQuerySchema.align()`` works without type coercion; also matches
+            MEDS's own ``DataSchema.code`` convention which uses ``large_string``
+            for the same 2 GB-offset reason).  Named ``query`` rather than ``code``
+            to match the column name already used throughout the sampler / dataset /
+            batch layer (``EveryQueryBatch.code`` is the distinct event-sequence-token
+            tensor).
         duration_days: The horizon, in days (continuous — ``float32``) within which the
             ``query`` code must occur for the query to be positive.  Allowing fractional
             days keeps the contract flexible for future finer-grained horizons.
@@ -84,3 +90,9 @@ class TaskQuerySchema(LabelSchema):
 
     query: Required(pa.large_string(), nullable=False)
     duration_days: Required(pa.float32(), nullable=False)
+    # Override ``boolean_value`` from ``LabelSchema`` (which declares it
+    # ``Optional(bool, nullable=NONE)``) to allow nulls — the EveryQuery task-label
+    # convention is to use null as the "censored" sentinel (closes #122).  The column
+    # stays optional at the schema level so inference-only inputs (no ground truth)
+    # continue to validate.
+    boolean_value: Optional(pa.bool_(), nullable=True)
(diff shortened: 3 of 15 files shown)


mmcdermott:
@copilot review

mmcdermott:
@copilot review

mmcdermott:
This is good, but a few questions:
1. Isn't this needed not just in predict and evaluate, but during training as well? And for sample tasks?
2. Do we need another schema as well -- one that extends this but adds labels, potentially for both occurs and censors?

mmcdermott:
Good questions — chewing on them, I think both have "yes" answers but the second one has a cleaner shape than I first reached for.

## 1. Should this be used by training + sample_tasks too?

**Yes.**  Currently the sampler hand-rolls its output schema in `sample_tasks.py` as `(subject_id, prediction_time, boolean_value, occurs, query, duration_days)`, and training's `EveryQueryPytorchDataset` consumes those column names directly (see `data/dataset.py:40` — `LABEL_TENSOR_NAMES = ("boolean_value", "censor", "occurs", "query", "duration_days")`).  Two drifts from TaskQuerySchema as filed here:

- Sampler uses `query` where this schema uses `code`.  Should reconcile to `code` (matches MEDS vocabulary convention).
- Sampler overloads `boolean_value` to mean *censored*, with a separate `occurs` column for "did the event actually fire."  This is the ad-hoc-ness your Q2 is pointing at.

I don't think this PR should take on the migration — TaskQuerySchema lands as the schema class, and a follow-on PR adopts it in sampler + training (with the `query` → `code` rename).  That keeps #96 narrow and the adoption diff reviewable on its own.

## 2. Do we need a labeled schema extending this?

**Yes — but the cleanest version is simpler than "another schema class."**  Two options, I lean toward Option B:

**Option A — dedicated `LabeledTaskQuerySchema`.**  Extends `TaskQuerySchema` with `occurs: Required(bool)` + `censored: Required(bool)`.  Keeps the current "occurs vs. censored" split explicit.  Then `PredictionSchema` (#99) extends `TaskQuerySchema` (not `LabeledTaskQuerySchema`) adding `predicted_boolean_probability`.

**Option B — just use MEDS's native convention on `boolean_value`.**  Drop the separate `occurs` + `censored` columns entirely.  `boolean_value = True/False` means *occurs / doesn't occur*.  `boolean_value = null` means *censored* (data ended before we could tell).  The loss mask becomes `boolean_value.is_not_null()` instead of `~batch.censor`.  This is the idiomatic MEDS shape — LabelSchema already treats `boolean_value` as nullable for exactly this case, and it removes the footgun where `boolean_value` and `occurs` both look like "did it happen?" but aren't (which burned me while writing the training-validity test in #119 — there's a big comment in that test explaining the gotcha).

Option B also makes `TaskQuerySchema` as-filed here already sufficient — no new schema needed.  Training-label input is just a `TaskQuerySchema` parquet with `boolean_value` filled in; inference input is the same parquet with `boolean_value` null (or omitted).  `PredictionSchema` stays as-is.

Option B is a bigger behavioral change (sampler and dataset have to rename `occurs` → `boolean_value` and flip the censor mask), but I think it's the right long-term shape.

## Suggested landing order

If you pick Option B: merge #96 as-is, file a follow-up issue for the sampler / dataset / model migration.  That work is sizable (sampler rewrite + dataset rewrite + model-loss rewrite) but incremental — each call site migrates independently.

If you pick Option A: I should add `LabeledTaskQuerySchema` to this PR (~15 LOC, low risk), then the same follow-up issue handles adoption.

Which direction?

mmcdermott:
Agreed to file the detailed migration doc — see #122 (step-by-step migration plan, every affected file, risks, test-guard story).

## On scope for this PR: I think you're right.  Bundle it.

My initial "keep narrow" instinct was "smaller PRs are easier to review."  But thinking harder, the argument cuts the other way here:

### Why the split would leave `dev` in a worse state than the bundled version

If #96 lands as just `TaskQuerySchema`, then `dev` has a cross-stage schema contract that **nothing uses**.  Everyone reading the code has to carry the mental footnote "we have TaskQuerySchema for the future, but the sampler's output still uses `query` + `occurs` + `boolean_value`-as-`censored`."  That's a documentation debt that compounds — every follow-up PR has to navigate the mismatch, and the dev-branch integration tests (#112 as restructured) would reference the old column names even after the schema class is defined.

The contract lands and nothing migrates is worse than either (a) contract + migration together, or (b) contract deferred until migration is ready.

### What bundling actually costs

From my audit in #122, the full migration is ~300-400 LOC across:
- `sample_tasks.py` (rename + rewrite `evaluate_index_df` label logic)
- `data/dataset.py` (drop `occurs` / `censor` from batch; loss mask derives from `boolean_value.is_not_null()`)
- `model/model.py` (loss mask flip; decide disposition of the `censor` prediction head)
- `test_sample_tasks.py` + `test_generate_tasks.py` + `test_training_validity.py` (test updates)
- CLI config files referencing `query.codes` — the `codes` field-name is already aligned; only internal struct references change

That's a big PR but not unwieldy, and each file's change is mechanical once the sampler's output shape is decided.  Review cost is actually lower bundled than split because the reviewer sees the full picture at once — no "where did `occurs` go?" moments.

### Critical dependency: #119

**The training-validity test in #119 is the strongest existing guard against the loss-mask flip.**  If `mask=~batch.censor` becomes `mask=batch.boolean_value.is_not_null()` and we silently mis-implement it, training loss still converges on garbage and we only find out at eval.  #119 would catch that in CI (HOT-COLD AUROC drops to chance).

Recommend this sequencing:

1. **Merge #119 first** on current (pre-Option-B) semantics — proves the test works against the existing training path.
2. **Expand #96** to full Option B adoption.
3. **Update #119** in #96 with the small test-only change (my synthetic labels use `boolean_value=False` for non-censored; under Option B that becomes `boolean_value=<occurs_truth>`, still non-null).  Confirm HOT-COLD AUROC still clears 0.8.
4. **Merge expanded #96.**

Step 1 gives us a reference point (known-good training under old semantics).  Step 4's green AUROC on the new semantics gives us strong evidence the migration didn't break training.

### Specific Option B decisions to lock before I start coding

- **Censor prediction head**: drop it entirely, or keep as optional aux loss behind a config flag?  I lean **drop** — "data availability" isn't a modeling target, and keeping the head means existing checkpoints still load without a state_dict surgery step.
- **Backward compat for existing checkpoints**: I lean **none**.  Existing trained `.ckpt` files will fail to load after the censor-head change; require a fresh train run.  The repo isn't published yet, blast radius is internal.  If I'm wrong about blast radius, we can keep the censor head trivially and just ignore its output — trivial to add back later.
- **`query` → `code` rename**: yes, as part of this PR.  Closes the TaskQuerySchema alignment gap at the same time.

Want me to proceed with expanding this PR along those lines?  If yes I'll start after #119 merges; if you want to also pull #119 into the same PR that's also defensible (makes the dependency explicit in the diff) but inflates the surface further.

mmcdermott:
Thanks for the corrections — I had three of them wrong.  Rewriting the plan:

## Corrected understanding of Option B

I was wrong to frame this as "drop the censor head + drop the occurs column."  The insight you're pointing at is cleaner and preserves both model heads:

**One `boolean_value` column encodes two conditional booleans.**  Because *occurs* is only meaningful conditional on *not censored*, a nullable boolean can carry both:

| `boolean_value` | Meaning |
|---|---|
| `null` | Censored — we can't answer the query for this patient |
| `True` | Not censored AND event occurred in the window |
| `False` | Not censored AND event did not occur |

The sampler used to write two columns (`boolean_value=censored`, `occurs=occurs_truth`) for what is semantically one conditional value; under Option B, one nullable column carries the same information.  **Both model heads stay** — `censor` predicts "can we answer for this patient?" and `occurs` predicts "conditioned on being able to answer, what would the answer be?"  Both are meaningful questions, both get supervised.

## Corrected migration

**Sampler** (`evaluate_index_df`):
- Write `boolean_value = None if censored else occurs_truth`.
- Drop the separate `occurs` column from the output parquet.

**Dataset** (`data/dataset.py::collate`):
- Derive both tensors from the one column:
  - `batch.censor = boolean_value.is_null()`
  - `batch.occurs = boolean_value` (with some placeholder like `False` where null, since masked out of loss)

**Model** (`model/model.py:614-621`): **unchanged**.  The existing loss wiring
```python
censor_loss = self._get_loss(censor_logits, batch.censor, mask=None)
occurs_loss = self._get_loss(occurs_logits, batch.occurs, mask=~batch.censor)
```
still does the right thing — it just consumes tensors derived from a different on-disk layout.

**Checkpoints**: stay loadable.  `state_dict` keys unchanged (both heads present).  Old `.ckpt` files work as-is.

This is a much smaller migration than I had sketched — the heavy lifting is `evaluate_index_df` + the dataset's `collate`, and nothing in the model.

## Corrected: not renaming `query` → `code`

Got it.  Then **`TaskQuerySchema` itself needs the column named `query`**, not `code`, to match the sampler.  Future generalisation (struct for ANY/ALL, event-bounded durations) can land as either a wider type (`query` as a struct) or a schema version bump, but using `code` now would bake in the false equivalence.

I'll rename the field in `schema.py` as part of the expanded #96 below.

## Updated scope for this PR

- `TaskQuerySchema` with `query: large_string` (rename from `code`) + `duration_days: float32`
- Migrate sampler's `evaluate_index_df` to single-column nullable `boolean_value`
- Migrate `data/dataset.py::collate` to derive `batch.censor` + `batch.occurs` from `boolean_value`
- Keep model unchanged
- Update affected unit/integration tests (`test_sample_tasks.py`, `test_generate_tasks.py` when rebased in)

Estimated scale: ~100-150 LOC, most in sampler + dataset.  No model code.  Backward-compat preserved for checkpoints.

## Updated #122 issue

I'll rewrite #122 to match the corrected understanding (both heads stay, no rename, backward-compat OK) so the migration doc doesn't mislead anyone who reads it after the fact.

## Sequencing — re: #119 review

You can **review #119 now, independently**.  It's on its own branch off dev, doesn't stack on #112 or #113, and tests the current semantics (two-column `boolean_value=censor, occurs=occurs_truth`) which is what's on dev today.  After #96 expands to include Option B, #119's hand-rolled labels will need a small test-only update (`boolean_value=False` everywhere becomes `boolean_value=occurs_truth`, still non-null since synthetic data is never censored), but that's trivial and doesn't affect the review of #119's substance now.

### Review order that probably makes most sense

1. **#119 now** — independent, validates current system works end-to-end.  Once green, gives you confidence that #96's migration hasn't broken training when it lands.
2. **#112** — independent of #119 but also off dev.  Fine to interleave with #119 review.
3. **#113** — stacks on #112; review after or in parallel.
4. **expanded #96** — after #119 is merged (so we have the safety-net test live on dev when the migration commit lands).

Want me to proceed with expanding #96 now (I'll rebase #119's test updates into it if it lands first, or add a trailing commit after #119 merges), or wait until #119 is in?

mmcdermott:
@copilot review

PR 96 scope expanded to wire `TaskQuerySchema`'s `code` column through the sampler + dataset (parquet-level rename `query` → `code`; internal `batch.query` / `self.query` attrs intentionally preserved — see commit message + inline comment in `data/dataset.py`).

Full change surface: 11 files, 87+/77-; all 203 non-slow tests pass locally.  Please review the rename coverage and the two-name convention rationale in `data/dataset.py`.

src/every_query/data/schema.py
```diff
@@ -0,0 +1,86 @@
...
+
+        >>> data = pa.Table.from_pylist([
+        ...     {"subject_id": 1, "prediction_time": datetime(2023, 1, 1),
+        ...      "query": "ICD//I10", "duration_days": 0.5},
+        ... ])
+        >>> _ = TaskQuerySchema.align(data)
+    """
+
+    query: Required(pa.large_string(), nullable=False)
+    duration_days: Required(pa.float32(), nullable=False)
```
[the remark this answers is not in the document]
mmcdermott: query should just be a string. It doesn't need to be a large string. You shouldn't need to write a custom align method.

src/every_query/data/dataset.py
```diff
@@ -332,10 +332,41 @@ def __init__(self, cfg: MEDSTorchDataConfig, split: str):
...
+        #     occurs: bool/long     ← boolean_value.fill_null(False) (masked on censored rows)
+        #
+        # Done here (after super().__init__() has already read the label parquet into
+        # self.schema_df) so the downstream getitem / collate path sees non-null tensors
+        # — torch's bool→tensor conversion doesn't tolerate nulls.
+        schema_cols = self.schema_df.collect_schema().names()
+        if "boolean_value" in schema_cols:
+            raw = pl.col("boolean_value")
+            self.schema_df = self.schema_df.with_columns(
+                raw.is_null().alias("boolean_value"),
```
mmcdermott, line 363: why call this boolean_value -- why not use "censored"? This is for our own code consumption, right? We don't need to call it boolean value, and censored would be clearer, right?

src/every_query/generate_tasks/sample_tasks.py
```diff
@@ -269,14 +273,13 @@ def evaluate_index_df(
         max_time_per_subject: Output of ``compute_max_time_per_subject``.
 
     Returns:
-        DataFrame with columns ``(subject_id, prediction_time, boolean_value, occurs, query,
-        duration_days)``.
+        DataFrame with columns ``(subject_id, prediction_time, boolean_value, query,
+        duration_days)``.  ``boolean_value`` is nullable (``null`` = censored).
     """
     out_schema = {
```
mmcdermott: you should be able to replace the use of out_schema entirely here with the schema type we define. That's the whole point. Look at how flexible_schema is used in other areas of the MEDS ecosystem. Ask me if you're confused.

tests/test_eval_suite.py
```diff
@@ -130,20 +130,30 @@ def test_invalid_type_raises(self):
 
 class TestGenTaskOutputSchema:
```
mmcdermott, line 132: use the schema you define to do this validation: https://github.com/medical-Event-Data-Standard/flexible_schema#table-and-schema-validation-and-alignment

tests/test_sample_tasks.py
```diff
@@ -805,12 +796,10 @@ def test_sampler_output_is_drop_in_for_dataset(self, tensorized_cohort_dir, tmp_
             "subject_id",
             "prediction_time",
             "boolean_value",
-            "occurs",
             "query",
             "duration_days",
         }
         assert labeled.schema["boolean_value"] == pl.Boolean
-        assert labeled.schema["occurs"] == pl.Boolean
```
mmcdermott, line 813: you should be able to replace this schema check logic with flexible_schema stuff as well

src/every_query/data/schema.py
```diff
@@ -0,0 +1,86 @@
...
+required query fields ``query`` and ``duration_days``.  Mirrors the pattern
+``meds-evaluation``'s `PredictionSchema
+<https://github.com/kamilest/meds-evaluation/blob/main/src/meds_evaluation/schema.py>`_ uses.
+
+The ``query`` column holds the MEDS code the query asks about.  The field is named
+``query`` (not ``code``) to match the existing column name used by ``sample_tasks``
+output, ``EveryQueryPytorchDataset``, and the ``EveryQueryBatch.query`` tensor —
+renaming to ``code`` would (a) collide with the inherited ``EveryQueryBatch.code``
+sequence-token tensor, which is a different thing semantically, and (b) churn every
+downstream consumer for no functional win.
```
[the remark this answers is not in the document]
mmcdermott: Fixed — PR description updated (see below).  Also expanded the module-level docstring in 261d856 to explain the naming choice: `query` matches what the sampler / dataset / batch layer already call this column, and `code` would collide with `EveryQueryBatch.code` (the inherited event-sequence-token tensor).

src/every_query/data/schema.py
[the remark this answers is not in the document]
mmcdermott: Took a different approach that avoids the override entirely — match the schema to what producers emit natively:

- polars' `pl.Utf8` serialises to `pa.large_string` (not `pa.string` — I confirmed via `df.to_arrow().schema`), so `query: Required(pa.large_string)` aligns natively.
- Sampler's `build_index_df` now emits `pl.Float32` for `duration_days` directly (was `pl.Int64`), matching `duration_days: Required(pa.float32)`.
- Evaluate's `gen_task.py` casts `pl.lit(duration)` to `pl.Float32` before writing.

Net: sampler + evaluate outputs pass `TaskQuerySchema.validate()` with zero coercion, and the new boundary check in `run_worker` (`TaskQuerySchema.validate(labeled.to_arrow())` before atomic-write) fails loudly on any future drift instead of quietly coercing.  Also added `boolean_value: Optional(bool, nullable=True)` override on `TaskQuerySchema` since MEDS `LabelSchema` declares it `nullable=NONE` but the EQ convention (from #122) uses null as the censored sentinel — without this, `validate` rejects the collapsed labels.

src/every_query/data/dataset.py
mmcdermott: Honest answer: the schema_df column is consumed by super (`MEDSPytorchDataset`), not just by our code, so we can't freely pick the name.  Specifically:

- `MEDSPytorchDataset.LABEL_COL = "boolean_value"` is class-level and wired through (a) parquet-read column selection (line 733 of super), (b) `self.labels = self.schema_df[self.LABEL_COL]` caching at super's `__init__` time, and (c) the per-item dict key used in `_seeded_getitem` / `collate`.
- Overriding `LABEL_COL = "censored"` on our subclass breaks super's parquet-read path (`if self.LABEL_COL in schema.names else ...` — our parquet has `boolean_value`, not `censored`), so super would conclude we have no task labels.
- The clean name **does** exist — just not in `schema_df`.  At the batch API layer `EveryQueryBatch.censor` is the explicit censor-indicator tensor that downstream code (model loss wiring, metrics, training-validity test) reads; `batch.boolean_value` (super-inherited) is effectively an alias for it post-collapse.

So the `boolean_value` naming is load-bearing for super's plumbing, not a stylistic choice.  I expanded the comment at the post-processing site in `dataset.py` to spell this out.  If we ever want to rename at the schema_df level, it'd be a separate refactor that subclasses more of `MEDSPytorchDataset` than feels worth it for a naming preference — happy to revisit once more of the pipeline is eval-aware.

src/every_query/generate_tasks/sample_tasks.py
mmcdermott: Done in 261d856.  The `evaluate_index_df` empty-input path now uses `TaskQuerySchema.schema()` directly to build an empty arrow table and projects to the 5-column label shape — no more hand-rolled polars-schema dict.  Also added `TaskQuerySchema.validate(labeled.to_arrow())` at the `run_worker` write boundary, so drift fails loudly before `_atomic_write_parquet`.  The hot-path compute (join_asof + censor-mask) still needs to touch polars expressions directly, but the schema is now the source of truth for what gets written out.

tests/test_eval_suite.py
mmcdermott: Done in 261d856.  `test_columns` → `test_conforms_to_task_query_schema`, which just calls `TaskQuerySchema.validate(df.to_arrow())`.  Drift in either dtype or column set now surfaces with a field-level diff directly from flexible_schema rather than a generic column-set mismatch.  Also dropped the redundant `test_non_censored_rows_are_non_null` / `test_no_null_occurs` which were tautologies under the new collapsed label.

tests/test_sample_tasks.py
mmcdermott: Done in 261d856.  Both `TestRunWorkerPipeline::test_writes_only_labels_parquet` and `TestEndToEndWithDataset::test_sampler_output_is_drop_in_for_dataset` now validate the labeled parquet via `TaskQuerySchema.validate(labels.to_arrow())` instead of hand-rolling the column set.  Also updated the test's hand-built index_df to use `pl.Float32` for `duration_days` so it matches the schema contract (the unit-level edge-case tests that don't exercise the write path can still use Int64).

src/every_query/generate_tasks/sample_tasks.py
mmcdermott, line 212: You can use the schema here, right? Even if not as a pyarrow directly, just to get the names and types of columns? https://github.com/medical-Event-Data-Standard/flexible_schema#exported-names-and-types

src/every_query/generate_tasks/sample_tasks.py
```diff
@@ -269,19 +280,20 @@ def evaluate_index_df(
...
+    # schema itself (rather than hand-rolling a matching polars schema dict).  The
+    # downstream consumer only cares that the column set + dtypes match the schema;
+    # using TaskQuerySchema.schema() directly eliminates the drift risk of keeping two
+    # parallel type lists in sync.
     if index_df.height == 0:
-        return pl.DataFrame(schema=out_schema)
+        arrow_schema = TaskQuerySchema.schema()
+        empty_table = pa.Table.from_pylist([], schema=arrow_schema)
+        return pl.from_arrow(empty_table).select(
+            "subject_id", "prediction_time", "boolean_value", "query", "duration_days"
```
mmcdermott: you can reference names from the class to avoid misalignment: https://github.com/medical-Event-Data-Standard/flexible_schema#exported-names-and-types

src/every_query/generate_tasks/sample_tasks.py
```diff
@@ -269,19 +280,20 @@ def evaluate_index_df(
...
-    }
+    # Empty-input fast-path: build an empty TaskQuerySchema-conformant table via the
+    # schema itself (rather than hand-rolling a matching polars schema dict).  The
+    # downstream consumer only cares that the column set + dtypes match the schema;
+    # using TaskQuerySchema.schema() directly eliminates the drift risk of keeping two
+    # parallel type lists in sync.
     if index_df.height == 0:
-        return pl.DataFrame(schema=out_schema)
+        arrow_schema = TaskQuerySchema.schema()
+        empty_table = pa.Table.from_pylist([], schema=arrow_schema)
```
mmcdermott: if you're doing this commonly enough, you could just have a helper to make an empty table given the schema, maybe?

src/every_query/generate_tasks/sample_tasks.py
```diff
@@ -324,13 +336,11 @@ def evaluate_index_df(
...
-        .with_columns(
-            pl.col("_censored").alias("boolean_value"),
-            (pl.col("_censored").not_() & event_in_window).alias("occurs"),
-        )
-        .select("subject_id", "prediction_time", "boolean_value", "occurs", "query", "duration_days")
+    # Collapsed nullable label: null when censored, else True/False from event_in_window.
+    boolean_value = pl.when(censored).then(pl.lit(None, dtype=pl.Boolean)).otherwise(event_in_window)
+
+    return joined.with_columns(boolean_value.alias("boolean_value")).select(
+        "subject_id", "prediction_time", "boolean_value", "query", "duration_days"
```
mmcdermott: ditto prior comments.

src/every_query/generate_tasks/sample_tasks.py
```diff
@@ -482,6 +492,13 @@ def run_worker(
     labeled = evaluate_index_df(index_df, events_df, max_time_df)
     # Downstream MEDS dataloader does not use task_id; it is intentionally absent from the
     # published schema.
+    #
+    # Validate the output against TaskQuerySchema at the write boundary.  The column set
+    # and types should already conform (evaluate_index_df emits schema-shaped output and
+    # duration_days is Float32 from build_index_df), so this is a guardrail against silent
+    # upstream drift rather than a coercion step — if ``align`` has to do any work we
+    # treat that as a bug in the producer, not something the write path papers over.
+    TaskQuerySchema.validate(labeled.to_arrow())
```
mmcdermott: There is an align method, that returns and might be more suited here, I think? Check flexible_schema and see. MEDS-transforms and MEDS-Extract also make use of this

src/every_query/predict/external_tasks/aces_to_eq.py
mmcdermott: do you want to validate against the schema here too?

src/every_query/generate_tasks/sample_tasks.py
mmcdermott: Done in 4bc0905. `evaluate_index_df` now uses the exported `{col}_name` attributes — `out_cols` is built from `TaskQuerySchema.subject_id_name`, `.prediction_time_name`, `.query_name`, `.duration_days_name` — and the empty branch delegates to the new `empty_task_query_df()` helper. Event-frame columns use `DataSchema.subject_id_name` / `.code_name` / `.time_name` from `meds.DataSchema` for the same reason.

src/every_query/generate_tasks/sample_tasks.py
mmcdermott: Done in 4bc0905. Every hardcoded `"subject_id"` / `"prediction_time"` / `"boolean_value"` / `"query"` / `"duration_days"` literal across `sample_tasks.py` is replaced with the schema's exported `{col}_name` attributes. Events-frame columns pull from `DataSchema.{col}_name` as well, so a rename on either schema flows through without edits.

src/every_query/generate_tasks/sample_tasks.py
mmcdermott: Done in 4bc0905. Added `empty_task_query_df()` in `every_query.data.schema` — returns an empty polars DataFrame keyed off the schema's exported `_name` attributes + `Float32` / `Boolean` dtypes. Exported from `every_query.data` and used from `evaluate_index_df`'s empty path.

One caveat worth noting in the helper's docstring: it's scoped to required columns + `boolean_value` (not every `LabelSchema` optional). Including `categorical_value` would blow up at `validate()` time because the polars→arrow round-trip coerces `pa.string` → `pa.large_string` on that column. The narrower shape is what downstream writers actually emit, which avoids the type-drift landmine.

src/every_query/generate_tasks/sample_tasks.py
mmcdermott: Done in 4bc0905 — applied the same schema-name-driven pattern to the rest of `sample_tasks.py` (all column selects, renames, and the `run_worker` label-frame construction).

src/every_query/data/dataset.py
```diff
@@ -332,10 +332,41 @@ def __init__(self, cfg: MEDSTorchDataConfig, split: str):
...
+                raw.fill_null(False).alias("occurs"),
+            )
+            schema_cols = self.schema_df.collect_schema().names()
+            # Super cached the pre-collapse ``boolean_value`` Series into ``self.labels``
+            # at its ``__init__`` time; re-point it at the post-processed (non-null,
+            # censor-indicator) column so ``_seeded_getitem`` and ``collate`` see the
+            # correct values.
```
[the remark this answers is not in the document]
mmcdermott: Done in 4bc0905. `EveryQueryPytorchDataset.__init__` now detects legacy two-column parquets by the presence of an `occurs` column before collapsing. The branch:

```python
schema_cols = self.schema_df.collect_schema().names()
is_legacy_two_col = "occurs" in schema_cols
if "boolean_value" in schema_cols and not is_legacy_two_col:
    # collapsed → legacy: null=censored, True=occurs
    ...
```

So legacy parquets (with both `occurs` and `boolean_value`) pass through untouched, and new collapsed parquets (nullable `boolean_value` only) are expanded to the legacy two-column form super-class consumers expect. Covered by existing sampler→dataset integration tests.

src/every_query/generate_tasks/sample_tasks.py
mmcdermott: Good catch — switched in 4bc0905. Did the flexible_schema deep-dive you suggested:

- `Schema.align()` returns the aligned arrow table with columns reordered + types coerced (e.g. polars' `Utf8` → `large_string`), whereas `validate()` is a strict check that errors on type mismatches.
- MEDS-transforms uses `align()` at write boundaries for exactly this reason; meds-testing-helpers does `DataSchema.align(df.to_arrow())` in its fixtures.

`run_worker` in `sample_tasks.py` and `create_eq_task_df` in `aces_to_eq.py` both now do:

```python
aligned = TaskQuerySchema.align(df.to_arrow())
pl.from_arrow(aligned).write_parquet(out_fp)
```

This lets producers emit their natural polars types without having to pre-coerce.

conftest.py
```diff
@@ -315,7 +326,6 @@ def task_labels_dir(tmp_path_factory: pytest.TempPathFactory) -> Path:
                 "prediction_time": pl.Datetime("us"),
                 "subject_id": pl.Int64,
                 "boolean_value": pl.Boolean,
-                "occurs": pl.Boolean,
                 "query": pl.Utf8,
                 "duration_days": pl.Int64,
             }
```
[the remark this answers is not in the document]
mmcdermott: Fixed in 4bc0905. The `task_labels_dir` fixture now stamps a true float `duration_days: 30.0` with `.cast(pl.Float32)`, so it aligns to `TaskQuerySchema.duration_days` (`pa.float32`) without a type coercion. Same fix applied to the `demo_dataset` fixture (which had been casting to `pl.Float64`).

src/every_query/predict/external_tasks/aces_to_eq.py
```diff
@@ -42,12 +42,20 @@ def create_eq_task_df(
...
             joined_df.select([*base_cols + code])
-            .rename({code: "occurs", "censored": "boolean_value"})
-            .with_columns(pl.lit(code).alias("query"))
-            .select(["subject_id", "prediction_time", "query", "boolean_value", "occurs", "task_label"])
+            .with_columns(
+                boolean_value.alias("boolean_value"),
+                pl.lit(code).alias("query"),
+            )
+            .select(["subject_id", "prediction_time", "query", "boolean_value", "task_label"])
         )
```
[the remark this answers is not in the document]
mmcdermott: Fixed in 4bc0905. `create_eq_task_df` now takes a `duration_days: float` parameter and stamps it into the output:

```python
pl.lit(duration_days).cast(pl.Float32).alias(TaskQuerySchema.duration_days_name)
```

The output select includes the full `TaskQuerySchema` column set (subject_id, prediction_time, query, duration_days, boolean_value) plus the bookkeeping `task_label`, and the write goes through `TaskQuerySchema.align(...)` at the boundary.

ACES tasks carry their horizon in the task config rather than per-row, so `duration_days` is a single constant per conversion run — the aces_to_eq.yaml now has `duration_days: 30.0` with a comment explaining how to override per task (`readmiss_30d` → 30, `mortality_1y` → 365, etc.).

src/every_query/predict/external_tasks/aces_to_eq.py
mmcdermott: Done in 4bc0905 via `TaskQuerySchema.align(...)` at the write boundary (same pattern as the sampler's `run_worker`). `align()` is actually a better fit than `validate()` here — it coerces the polars-native types (e.g. `Utf8` → `large_string`) rather than forcing the caller to pre-coerce — and if any required column is missing or a type is unreconcilable, it still errors, so the contract is upheld.

src/every_query/generate_tasks/sample_tasks.py
```diff
@@ -482,7 +509,15 @@ def run_worker(
...
-    _atomic_write_parquet(labeled, labels_fp)
+    #
+    # Align the output against TaskQuerySchema at the write boundary.  ``align`` is
+    # preferred over ``validate`` here because it (a) reorders columns into the schema's
+    # canonical order and (b) casts mistyped columns to the declared dtypes — so even if
+    # an upstream edit silently changes one type (e.g., reverting ``build_index_df`` to
+    # emit ``Int64`` durations), the on-disk parquet still conforms.  Missing or extra
+    # columns remain hard errors (``align`` re-raises on those).
+    aligned = TaskQuerySchema.align(labeled.to_arrow())
+    _atomic_write_parquet(pl.from_arrow(aligned), labels_fp)
```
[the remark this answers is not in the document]
mmcdermott: Good catch on the doc/code mismatch. Updated the PR description to match the implementation — `run_worker` does call `align()`, not `validate()`, and the description now explains why (polars emits `Utf8` which arrow-izes to `large_string`; `align()` handles the reorder/coerce while still erroring on missing/extra columns or unreconcilable type drift, matching the MEDS-transforms / meds-testing-helpers pattern).

This is intentional coercion, not drift-masking — the schema pins `large_string` specifically *because* that's what the producer's polars→arrow round-trip emits, so there's no silent type change in practice.

src/every_query/predict/external_tasks/aces_to_eq.py
```diff
@@ -19,8 +20,21 @@ def shard_id(p: Path) -> int:
...
 
 def create_eq_task_df(
-    eq_shard_fp: str, aces_shard_fp: str, codes: list[str], output_root: str, target_rows: int
+    eq_shard_fp: str,
+    aces_shard_fp: str,
+    codes: list[str],
+    output_root: str,
+    target_rows: int,
+    duration_days: float,
 ) -> int:
```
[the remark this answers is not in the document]
mmcdermott: Fair observation, but the unused `target_rows` parameter is pre-existing scaffolding (landed in 4d7f974 when `aces_to_eq/` was relocated under `predict/external_tasks/`) — not introduced by this PR. This PR's scope is the TaskQuerySchema refactor + collapsed nullable label; removing dead-parameter dusting from the aces pipeline is orthogonal.

I'll file a follow-up issue to either wire `target_rows` into a real sampling step or remove it from the signature + config. Keeping this PR focused.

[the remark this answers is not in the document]
mmcdermott: Filed #130 to track.

PR #97, opened 2026-04-18
refactor(paper_experiments): make sample_codes dataset-agnostic (closes #85)

### LINKED ISSUE #85: Phase 4.2: Rewrite sample_codes/ to be dataset-agnostic, move to paper_experiments/
Parent: #54 (Phase 4.2). Depends on: #84.

## Goal

Rewrite the scripts under `src/every_query/sample_codes/` to work on any MEDS cohort — no hardcoded MIMIC paths — and relocate them under `src/every_query/paper_experiments/sample_codes/`.

Per [Matthew's 2026-04-18 comment on #54](https://github.com/payalchandak/EveryQuery/issues/54#issuecomment-TBD): core EQ accepts a code list as input (users pick codes of interest and pass them in); *generating* that code list — especially with ID/OOD splits — is a paper-experiments construct. Normal usage doesn't hold codes out.

## What's broken today

`src/every_query/sample_codes/sample_train_codes.py:10`:

```python
PARQUET_PATH = "/users/gbk2114/data/MIMIC_MEDS/MEDS_cohort/processed/metadata/codes.parquet"
```

Nobody outside that cluster can run it. Similarly hardcoded paths in `sample_eval_codes.py`, `sample_embedding_codes.py` (verify at implementation time).

## Proposed shape

1. Move `src/every_query/sample_codes/*` → `src/every_query/paper_experiments/sample_codes/*`.
2. Rewrite each script to take `--metadata-dir` (or Hydra `metadata_dir`) pointing at any MEDS cohort's `{metadata_dir}/codes.parquet`.
3. Parametrize filter thresholds (min subject count, etc.) via Hydra config rather than constants.
4. Register `EQ_sample_codes` (or equivalent) entry point — or keep as `python -m every_query.paper_experiments.sample_codes.sample_train_codes` if a CLI feels heavyweight for a paper-experiment script.

## Tests

- Unit test: run the sampler against a tiny synthetic `codes.parquet`, assert the output has the expected shape.
- No integration test needed — it's a dataset utility.

## Scope exclusions

- No changes to how codes are consumed by `EQ_pretrain` or downstream — that interface stays the same (a list of codes in, via Hydra).
- ID/OOD split semantics stay unchanged; just become path-agnostic.

## Acceptance criteria

- [ ] `sample_codes/` lives under `paper_experiments/`.
- [ ] No hardcoded paths remain; all inputs are Hydra-configurable.
- [ ] At least one smoke test exercising the path-agnostic flow.
- [ ] Top-level README + `paper_experiments/README.md` updated to reference the new location.

mmcdermott:
## Summary

Per @mmcdermott on #54: these scripts shouldn't be tied to one specific cluster path — *"this should just point to a MEDS dataset, then pick a set of codes to include based on the metadata/codes.parquet file, then write a list out to a file on disk that you can use. I don't see a reason why there should be any hard coded MIMIC paths."*

Closes #85.

## What changes

- `PARQUET_PATH = "/users/gbk2114/data/MIMIC_MEDS/..."` → `--metadata-dir` CLI arg on all three scripts.
- `OUT_DIR = "../train_codes"` (relative, cwd-dependent) → `--out-dir` CLI arg.
- Filter thresholds (`N_SAMPLES`, `N_REPEATS`, `SEED`, and the previously-implicit `TIME` exclude) all become CLI flags with sensible defaults.
- Shared `load codes.parquet + filter TIME` step and `stable_hash_list` helper factored into `_common.py` so a future schema change to `codes.parquet` is one edit.

No sampling semantics change — the three scripts do exactly what they did before, just with configurable inputs.

## Entry points — intentionally *not* registered

Sampler scripts are invoked a couple of times per paper run, not production surface. Leaving them as `python -m every_query.paper_experiments.sample_codes.<script>` keeps the top-level `EQ_*` namespace focused on the normal-usage pipeline. Easy one-line addition to `pyproject.toml` if that judgment changes.

## Usage example

```bash
python -m every_query.paper_experiments.sample_codes.sample_train_codes \
    --metadata-dir "$PROCESSED/metadata" \
    --out-dir "$PROJECT_DIR/train_codes" \
    --n-samples 10000 \
    --n-repeats 5
```

## Test plan

- [x] `stable_hash_list` doctest (determinism + order-sensitivity + configurable length) passes.
- [x] `pytest tests/ src/` — 192 passed.
- [x] Manually running `python -m every_query.paper_experiments.sample_codes.sample_train_codes --help` works on the packaged install.

### DIFF
--- a/src/every_query/paper_experiments/sample_codes/__init__.py
+++ b/src/every_query/paper_experiments/sample_codes/__init__.py
@@ -1,0 +1,11 @@
+"""Query-code sampling utilities for the EveryQuery paper experiments.
+
+Three argparse CLIs — see each module's docstring for details:
+
+- ``sample_train_codes`` — sample N-code YAMLs for PT runs.
+- ``sample_eval_codes`` — sample paired ID/OOD eval YAMLs given an existing train YAML.
+- ``sample_embedding_codes`` — sample codes for post-hoc embedding-visualization figures.
+
+All three load their code universe from ``{metadata_dir}/codes.parquet`` passed on the CLI —
+no hardcoded paths.  ``_common.py`` holds the shared filtering + hashing helpers.
+"""
--- a/src/every_query/paper_experiments/sample_codes/_common.py
+++ b/src/every_query/paper_experiments/sample_codes/_common.py
@@ -1,0 +1,73 @@
+"""Shared helpers for the sample-codes scripts.
+
+All three scripts in this submodule start from the same inputs — load a MEDS cohort's
+``metadata/codes.parquet``, optionally filter out codes matching an exclude pattern (TIME
+tokens by default) — and differ only in how they sample from the resulting code universe.
+Keeps those steps in one place so a future schema change to ``codes.parquet`` is a one-line
+update.
+"""
+
+import hashlib
+from pathlib import Path
+
+import polars as pl
+
+
+def stable_hash_list(items: list[str], length: int = 12) -> str:
+    """Order-sensitive, deterministic hash for a list of strings.
+
+    Used to name output files by content so re-running with the same inputs produces an
+    identically-named file (idempotent sampling run).
+
+    Examples:
+        Same list → same hash:
+
+        >>> stable_hash_list(["A", "B"]) == stable_hash_list(["A", "B"])
+        True
+
+        Order matters:
+
+        >>> stable_hash_list(["A", "B"]) != stable_hash_list(["B", "A"])
+        True
+
+        Length is configurable:
+
+        >>> len(stable_hash_list(["A"], length=6))
+        6
+    """
+    h = hashlib.sha256()
+    for x in items:
+        h.update(x.encode("utf-8"))
+        h.update(b"\n")
+    return h.hexdigest()[:length]
+
+
+def load_filtered_codes(metadata_dir: Path, exclude_pattern: str | None = "TIME") -> list[str]:
+    """Load the code universe from ``{metadata_dir}/codes.parquet`` and optionally filter.
+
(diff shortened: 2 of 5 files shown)


mmcdermott:
@copilot review

mmcdermott:
Addressed the two observations in e0ca660:

**(2) Bounds check before `random.sample`** — Applied to all three scripts. Each now calls `parser.error(...)` with a human-readable message naming the requested count and the available count *before* reaching `random.sample` (which would otherwise raise the cryptic `ValueError: Sample larger than population`).

**(1) File-open-mode inconsistency** — Pushing back on this one. The difference is intentional:

- `sample_train_codes` / `sample_embedding_codes` write to **content-hashed filenames** (`{N}_ID__{hash}.yaml`, `embed_{N}_{hash}.yaml`). Re-running with the same inputs produces the same path — exclusive-create (`"x"`) is the right idempotency guard; re-running with different inputs produces a different hash, different path, no collision.
- `sample_eval_codes` writes to a **user-supplied `--out-fp`**. The user has explicitly named the destination — overwriting it on rerun is the expected behavior (common pattern: tweak `--num-id-codes` and re-run into the same output path).

The two shapes are a feature, not a bug. Happy to add `--force` to `sample_train_codes` / `sample_embedding_codes` if we want an escape hatch, but that's scope creep for this PR.

mmcdermott:
@copilot review

mmcdermott:
@copilot review

mmcdermott:
This is probably ready to go, but I want to leave this alone for now and focus on the other parts of the repo first, before addressing this.

Issue #81, opened 2026-04-18
Phase 2.2: Add EQ_predict inference-only entry point


mmcdermott:
Parent: #54 (Phase 2.2). Depends on: #80 (schema), #79 (submodule restructure).

## Goal

Add an inference-only `EQ_predict` entry point that takes `(trained_model_run_dir, tasks.parquet)` and writes prediction parquets. Fully decoupled from metrics, model selection, or AUC computation.

## Why split this out

Today `EQ_evaluate` conflates four concerns: index-time sampling, task-matrix slicing, inference, and metric computation. Separating inference lets external consumers run the model on their own tasks without inheriting our metric conventions, and gives `EQ_evaluate` a clean input (predictions) it can compute metrics over.

## Scope

- New submodule: `src/every_query/predict/`
- New Hydra entry point: `EQ_predict`
- Input: `(model_run_dir, tasks_parquet_matching_schema)` where the schema is whatever #80 lands.
- Output: prediction parquets. Shape: the input rows plus a `predicted_boolean_probability` column (name TBD to match MEDS / meds-evaluation conventions).
- One model per invocation. Resolves the "why a list of model_run_dirs?" question from #66 by making the list go away.

## Explicitly excluded

- No AUCs. No ROC/PR. No per-bucket aggregation. No win-rate ranking. No model selection.
- No ID/OOD splitting. The caller provides the tasks parquet.
- No checkpoint selection policy — `model_run_dir` is expected to already resolve to one checkpoint (e.g. `best_model.ckpt`).

## Open questions

- Batch / streaming write — does prediction write one-parquet-per-shard, one-parquet-total, or follow the input's layout?
- Does `EQ_predict` load the model's own training config from `resolved_config.yaml` to recover datamodule settings, or does it accept an explicit datamodule config? (First is nicer, second is more flexible.)
- How do we handle codes present in the tasks parquet but absent from the model's training vocabulary? Error, warn, or skip?

## Acceptance criteria

- [ ] `EQ_predict --help` works after `pip install`.
- [ ] Given a trained demo model + a tasks parquet, produces a predictions parquet with the agreed schema.
- [ ] CPU-only integration test in `tests/test_predict_cli.py` that exercises the full subprocess path.

## Dependencies

- #80 (schema) must be decided before this starts implementation. Until then this issue is a placeholder.
- #79 (submodule restructure) — `EQ_predict` will live under `predict/` alongside the restructured `prepare_tasks/` and `pretrain/`.

[...]

Issue #68, opened 2026-04-17
CLI test upgrade: wheel-install validation + staged functional tests (beyond --help)


mmcdermott:
## Context

\`tests/test_cli_smoke.py\` (added in #61) covers exit-0 on \`EQ_* --help\` for all 7 endpoints. Two gaps worth closing, in staged order. Proposal informed by a read of MEDS_EIC_AR's CLI testing setup.

## Gap 1: Editable-install blind spot

**Concern.** The tests run against the uv-managed editable install (PATH-prepended \`.venv/bin\`). A user running \`pip install EveryQuery\` on a fresh machine could hit:
- Missing package-data globs (YAMLs dropped from the wheel) — the \`importlib.resources.files()\` paths from #61 would fail.
- Malformed \`[project.scripts]\` metadata.
- Missing transitive runtime deps (something in \`[dependency-groups].dev\` actually needed by runtime).

**Does MEDS_EIC_AR test this?** No — they also run against the editable install (\`uv sync --locked --group dev\` in \`.github/actions/setup\`), so we're at parity. But it's cheap to close the gap.

**Proposal — add a small CI-only job:**
\`\`\`yaml
- name: Wheel-install smoke
  run: |
    uv build
    uv venv /tmp/eq-wheel
    VIRTUAL_ENV=/tmp/eq-wheel uv pip install dist/*.whl
    for s in EQ_train EQ_evaluate EQ_generate_tasks EQ_generate_tasks_exhaustive \\
             EQ_gen_eval_index EQ_gen_eval_tasks EQ_select_model; do
      /tmp/eq-wheel/bin/\$s --help  # plus --config-dir smoke injection for EQ_{train,evaluate,gen_eval_tasks}
    done
\`\`\`

Runs once per commit, &lt;2 min, catches all packaging regressions that editable installs hide. Doesn't need to be a pytest test — a shell step in CI is sufficient.

## Gap 2: \`--help\` doesn't prove the CLIs actually run

**Concern.** Help works; a single override typo in a config could still blow up on first real invocation. MEDS_EIC_AR solves this with session-scoped fixtures that invoke the installed CLI as subprocess on synthetic MEDS data, then cascade downstream tests off the produced artifacts.

**Reference patterns (MEDS_EIC_AR):**
- \`/tmp/MEDS_EIC_AR/conftest.py:21-43\` — \`run_and_check()\` subprocess helper.
[...]

PR #95, opened 2026-04-18
feat: register EQ_aces_to_eq + EQ_process_composite entry points (closes #62)

mmcdermott:
## Summary

File moves landed in #90 — this PR registers the two Hydra mains as console scripts so they're directly invocable post-install.

Closes #62.

## What lands

- `pyproject.toml`: adds
  ```toml
  EQ_aces_to_eq        = "every_query.predict.external_tasks.aces_to_eq:main"
  EQ_process_composite = "every_query.predict.external_tasks.process_composite:main"
  ```
- `tests/test_cli_smoke.py`: parametrize list gains both entries. The `aces_to_eq` config defaults on a `query_codes` compose-group, so the smoke-config-dir fixture now also writes a `query_codes/smoke.yaml` stub alongside the existing `train_codes` / `eval_codes` stubs.

## Why not `get_per_code_from_composite` too

[@gkondas on #62](https://github.com/payalchandak/EveryQuery/issues/62#issuecomment-4269309300) only mentioned `aces_to_eq` and `process_composite`. `get_per_code_from_composite` is a helper used *after* composite aggregation to evaluate each code individually; the primary surface is the two registered here. If it needs an entry point later, it's a one-line addition.

## Test plan

- [x] `pytest tests/test_cli_smoke.py` — 9 passed (7 existing + 2 new).
- [x] `EQ_aces_to_eq --help` and `EQ_process_composite --help` both exit 0 after `uv sync`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #98, opened 2026-04-18
ci: wheel-install smoke — validate packaged entry points on every push (refs #68)

mmcdermott:
## Summary

Part 1 of #68 — the wheel-install smoke step.  Adds a `Wheel-install smoke` job to `python-build.yaml` that depends on the existing `build` job, downloads the wheel artifact, installs it into a fresh venv (no source tree), and runs `--help` on every registered `EQ_*` console script.

Refs #68.  Part 2 of that issue — staged functional tests via `@pytest.mark.slow` — is not in this PR; best landed after Phase 2 (#81 + #83) stabilizes the post-refactor entry-point surface.

## What this catches

- **Missing package-data YAMLs.**  Editable install sees everything in the source tree, so `importlib.resources.files("every_query") / "..." / "configs" / "foo.yaml"` always resolves.  A wheel install only has what `[tool.setuptools.package-data]` put into the wheel.  If anyone adds a new YAML and forgets to match the glob, this CI step fails.
- **Malformed `[project.scripts]` entries.**  A typo in a module path is silent on editable install (the entry is still *present*, the generated shim just blows up at invocation) but a wheel install round-trips the metadata — a malformed entry fails here.
- **Accidentally-runtime dev deps.**  If any runtime module imports something only listed in `[dependency-groups].dev`, editable install never notices; the wheel install does.

## How the step works

After the existing `build` job uploads `dist/*.whl` as the `python-package-distributions` artifact, the new job:

1. Downloads the artifact on a fresh runner.
2. Creates `/tmp/wheel-venv` via `uv venv`.
3. `uv pip install dist/*.whl` into that venv.
4. Stubs the compose-group YAMLs `EQ_train` / `EQ_evaluate` / `EQ_gen_eval_tasks` / `EQ_aces_to_eq` need (`train_codes/smoke.yaml`, `eval_codes/smoke.yaml`, `query_codes/smoke.yaml`) — mirrors what `tests/test_cli_smoke.py` already does.
5. Runs `--help` on each of the 9 registered `EQ_*` CLIs (5 without a compose stub, 4 with).
6. `set -euo pipefail` — any non-zero return fails the job.

## Test plan

- [x] CI run on this PR will exercise the new job; expect it to pass since it mirrors the editable-install smoke tests that already pass.
- [x] `pre-commit run --files .github/workflows/python-build.yaml` — clean (prettier doesn't touch, codespell OK).

### DIFF
--- a/.github/workflows/python-build.yaml
+++ b/.github/workflows/python-build.yaml
@@ -78,24 +78,26 @@ jobs:
           WANDB_ENTITY: test
         run: |
           set -euo pipefail
-          # Compose-group stubs for CLIs whose shipped configs default on gitignored YAMLs.
+          # Compose-group stubs for CLIs whose shipped configs default on gitignored YAMLs
+          # (see tests/test_cli_smoke.py for the source-of-truth pattern).
           SMOKE=/tmp/eq-smoke-cfg
-          mkdir -p "$SMOKE/train_codes" "$SMOKE/eval_codes" "$SMOKE/query_codes"
+          mkdir -p "$SMOKE/train_codes" "$SMOKE/eval_codes"
           printf 'codes: []\n' > "$SMOKE/train_codes/smoke.yaml"
           printf 'id: []\nood: []\n' > "$SMOKE/eval_codes/smoke.yaml"
-          printf -- '- A\n- B\n' > "$SMOKE/query_codes/smoke.yaml"
 
           BIN=/tmp/wheel-venv/bin
+
           # CLIs that don't need a compose stub.
-          for cli in EQ_process_data EQ_generate_tasks EQ_gen_eval_index EQ_select_model EQ_process_composite; do
+          for cli in EQ_process_data EQ_generate_tasks EQ_gen_eval_index EQ_select_model; do
             echo "=== $cli --help ==="
             "$BIN/$cli" --help > /dev/null
           done
+
           # CLIs that need a compose stub injected via --config-dir.
-          "$BIN/EQ_train"         --config-dir="$SMOKE" train_codes=smoke --help > /dev/null
-          "$BIN/EQ_evaluate"      --config-dir="$SMOKE" eval_codes=smoke --help > /dev/null
+          "$BIN/EQ_train" --config-dir="$SMOKE" train_codes=smoke --help > /dev/null
+          "$BIN/EQ_evaluate" --config-dir="$SMOKE" eval_codes=smoke --help > /dev/null
           "$BIN/EQ_gen_eval_tasks" --config-dir="$SMOKE" eval_codes=smoke --help > /dev/null
-          "$BIN/EQ_aces_to_eq"    --config-dir="$SMOKE" query_codes=smoke --help > /dev/null
+
           echo "All entry points resolve from the wheel install."
 
   publish-to-pypi:
--- a/.github/workflows/python-build.yaml
+++ b/.github/workflows/python-build.yaml
@@ -42,8 +42,6 @@ jobs:
       - build
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
-
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
@@ -74,7 +72,6 @@ jobs:
           PROCESSED: /tmp
           INTERMEDIATE: /tmp
           FINAL_DATA_DIR: /tmp
-          ACES_SHARDS_DIR: /tmp
           WANDB_ENTITY: test
         run: |
           set -euo pipefail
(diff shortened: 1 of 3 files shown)


mmcdermott:
@copilot review

mmcdermott:
Thanks for flagging the missing entry points — the root cause is that this PR was branched before #95 registered `EQ_aces_to_eq` / `EQ_process_composite`.  Copilot's fix `fa53863` is correct in substance but duplicates what #95 is already landing, which would produce a merge conflict.

Reverted + trimmed in `f902db9`: the smoke job now only exercises the 7 entry points present on `dev`.  Once #95 merges, a subsequent dev-merge into this branch will pick the two new ones up automatically — I'll add them back to the smoke loop in that merge.

mmcdermott:
@copilot review

mmcdermott:
Comments from chatgpt:
My take: this PR is **reasonable as a narrow phase-1 guardrail**, but **not the right end state** for CLI testing. It closes a real gap the repo already identified: editable-install tests can miss missing package data, broken entry-point metadata, and runtime deps that only fail after a real wheel install. That is exactly what issue #68 says the “wheel-install smoke” step is meant to catch. 

Where I think your discomfort is justified is that this PR is **only packaging smoke**, not “installed CLI integration testing” in any substantive sense. The new workflow builds the wheel, installs it into a fresh venv, and runs `--help` on the installed `EQ_*` scripts. That is useful, but it only proves “entry point resolves, imports don’t explode immediately, config discovery gets far enough for help text.” It does **not** prove a real command invocation works after install. 

There are three concrete problems with the current shape.

First, it **duplicates test logic in YAML**. The workflow recreates the temporary `train_codes/smoke.yaml` and `eval_codes/smoke.yaml` setup inline, and that same pattern already exists in `tests/test_cli_smoke.py`. That gives you two sources of truth for the same smoke contract. Drift is likely.  

Second, it is partly **testing around a design smell instead of forcing it into the open**. The existing smoke test explicitly says `EQ_train`, `EQ_evaluate`, and `EQ_gen_eval_tasks` need a temporary `--config-dir` because their canonical defaults are generated out-of-band and not checked in. That is brittle for an installed CLI surface. Either those defaults are part of the public packaged interface and should be shipped, or the CLI should require explicit config instead of pretending a default exists. Right now the PR preserves that ambiguity. 

Third, there is some **evidence of drift already**. The PR description says it runs help for “9 registered `EQ_*` console scripts,” but the current `pyproject.toml` defines 7 scripts, and the workflow itself exercises 7. That is not a fatal bug, but it suggests the PR is a bit more ad hoc than I’d want for test infrastructure.   

So the answer to “can’t this be the right way?” is:

* **As a cheap sentinel:** yes, it is fine.
* **As the actual solution:** no, it is incomplete and slightly papering over the real issue.

The more important issue is how to verify that CLI tests work **after a pip install**, not just in editable mode.

The key rule is: **do not run the installed-mode tests from the repo root**. If you do, Python can still see the checkout, and local imports can mask packaging problems. The current train integration test already runs a real subprocess, but it uses `python -m every_query.train.train` from the dev environment, so it still does not validate the installed console script surface or a wheel-installed import graph. 

The robust pattern is:

1. Build the wheel.
2. Create a fresh venv.
3. Install the wheel into that venv.
4. Run a small set of **black-box installed tests from outside the checkout**.
5. Invoke only the installed console scripts or the installed interpreter from that venv.

Concretely, I would do something like this in CI:

```bash
uv build

uv venv /tmp/eq-installed
VIRTUAL_ENV=/tmp/eq-installed uv pip install dist/*.whl pytest meds_testing_helpers

mkdir -p /tmp/eq-installed-tests
cp -r tests_installed /tmp/eq-installed-tests/

cd /tmp/eq-installed-tests
/tmp/eq-installed/bin/pytest -q
```

And then make `tests_installed/` contain only black-box tests. For example:

```python
# tests_installed/test_train_installed.py
import os
import subprocess
import sys
from pathlib import Path

def test_eq_train_installed_runs(tmp_path):
    bin_dir = Path(sys.executable).parent
    env = os.environ.copy()

    for var in (
        "PROJECT_DIR", "OUTPUT_DIR", "TASK_DIR",
        "PROCESSED", "INTERMEDIATE", "FINAL_DATA_DIR",
    ):
        env.setdefault(var, str(tmp_path))
    env.setdefault("WANDB_ENTITY", "test")

    # create tiny synthetic inputs here, or stage them in tmp_path
    # then call the installed console script, not python -m from the checkout
    cmd = [
        str(bin_dir / "EQ_train"),
        "--config-name=_demo_train",
        f"output_dir={tmp_path / 'out'}",
        # other overrides...
    ]
    result = subprocess.run(cmd, capture_output=True, text=True, env=env, timeout=180)
    assert result.returncode == 0, result.stderr
```

That catches the failure mode you actually care about:

* wrong `[project.scripts]`
* missing packaged YAML/config assets
* missing runtime deps
* code that only works because the checkout is on `sys.path`
* CLIs that only work in dev/editable environments

My recommendation for this repo specifically would be:

* **Keep PR #98 or a version of it**, because it is cheap and useful.
* But **move the smoke harness out of the workflow YAML** into a committed script or a pytest target.
* Then add a separate `installed` or `wheel` test job that runs **real black-box CLI invocations from a temp directory outside the repo**.
* At minimum, cover one real invocation each for:

  * training
  * task generation
  * the eval chain
* And decide explicitly whether those out-of-band default YAML groups are part of the shipped CLI contract or not. The tests should enforce that decision rather than work around it.  

So overall: **the PR is not wrong, but it is only a first rung on the ladder**. Your instinct is right if the concern is “this shouldn’t be the main answer to installed-mode CLI correctness.”

mmcdermott:
Overall, though, this is overkill for us for now. Closing as not planned.

Issue #82, opened 2026-04-18
Phase 2.3: Inventory eval_suite + eval.py code paths, classify each operation


mmcdermott:
Parent: #54 (Phase 2.3). Prerequisite for: #83.

## Goal

Walk every operation in the current eval pipeline and classify it. This is the **discovery step** that produces a migration plan; no code changes.

## Files in scope

- `src/every_query/eval.py`
- `src/every_query/eval_suite/gen_index_times.py`
- `src/every_query/eval_suite/gen_task.py`
- `src/every_query/eval_suite/select_model.py`
- Their configs under `src/every_query/eval_suite/conf/`

Plus any helper utilities they pull in (e.g. `utils/codes.py::code_slug`).

## Classification axes

For each logical operation inside the four files, assign one of:

| Bucket | Destination post-refactor | Examples |
|---|---|---|
| **Core inference** | `predict/` (#81) | Model load, forward pass, probability write |
| **Core metrics** | `evaluate/` (#83) | AUROC, AUPRC per (code, duration), rolling CI |
| **Paper experiments** | `paper_experiments/` (#84) | Pairwise win-rate ranking, ID-vs-OOD bucket reporting, cross-model comparison tables |
| **Obsolete** | Delete | Anything made redundant by the new schema |

## Key questions the inventory should answer

1. **`gen_index_times`**: does its "sample K prediction times per subject at eval time" logic survive? If yes, under `prepare_tasks/` (PT and eval both need it)? Under `evaluate/`?
2. **`gen_task`**: once predictions come from a schema-conformant tasks parquet (#81), does the per-(code, duration) slicing here have any remaining purpose? Suspicion: no — the caller just writes the tasks they want and `EQ_predict` handles it.
3. **`select_model`**: cross-model pairwise win-rate is a *ranking* operation. Is that core evaluation, or is it specifically for our paper's Figure N comparing trained variants? Matthew's 2026-04-18 comment suggests the latter.
4. **`eval.py::_run_test`**: the bucket-label-by-ID-vs-OOD behavior (#32) — is ID/OOD labeling a core-metrics operation, or is the split itself paper-experiments only?
5. **`model_run_dirs` as a list** (#66): does any surviving component still need a list of models, or does that always live in paper-experiments / leaderboard code?

## Deliverable

[...]

PR #99, opened 2026-04-18
feat(predict): EQ_predict inference-only entry point (closes #81)

### LINKED ISSUE #81: Phase 2.2: Add EQ_predict inference-only entry point

mmcdermott:
## Summary

Closes #81.  Implements `EQ_predict` — an inference-only entry point that takes a trained-model run directory + a directory of `TaskQuerySchema`-conformant parquets and produces a `PredictionSchema`-conformant parquet.

## What lands

- **`predict/schema.py`** — `PredictionSchema(TaskQuerySchema)` adds `censor_prob: Required(pa.float32(), nullable=False)` + `occurs_prob: Required(pa.float32(), nullable=False)`.  Two columns rather than one because the model has two heads; emits both independently so downstream evaluation can compute conditional metrics.  #129 tracks the post-refactor `occurs_prob → label_prob` generalization.

- **`predict/predict.py`** — `EQ_predict` Hydra main.  Single, order-preserving pass:
  1. Validate `tasks_dir` (must be a directory of `.parquet` files; warn on >1 parquet, error on non-parquet files or empty dir).
  2. Load the trained model via `every_query.utils.model_loader.setup_model(model_run_dir, ckpt_name)` — resolves `best_model.ckpt` → `checkpoints/last.ckpt` → explicit ckpt.
  3. Instantiate the datamodule with `task_labels_dir = tasks_dir`.  Pick dataset + dataloader per the `split` config (`held_out` default; `tuning` supported for intermediate-validation uses like the training-validity E2E).
  4. Pre-flight `_warn_out_of_vocab` — task codes missing from the model's training vocab PAD-encode silently; the warn surfaces that without hard-erroring.
  5. `trainer.predict(dataloaders=...)` on the whole flat frame.
  6. Build the output identifier columns from the dataset's `schema_df` (MTD guarantees order+length preservation); reconstruct the nullable `boolean_value` from the collapsed `(censor, occurs)` pair (inverse of `EveryQueryPytorchDataset.__init__`'s collapse).
  7. `hstack` probabilities onto identifiers (no key-based join; row-count check guards the MTD invariant), align to `PredictionSchema`, write the aligned arrow table directly via `pq.write_table`.

- **`predict/configs/predict.yaml`** — required: `model_run_dir`, `tasks_dir`, `output_parquet`.  Optional: `ckpt_name` (defaults to `best_model.ckpt`/`last.ckpt`), `split` (`held_out` default, also accepts `tuning`).

- **`utils/model_loader.py`** — `setup_model` extracted from `evaluate/eval.py` to a shared utility so `EQ_predict` doesn't couple to a module slated for deletion in the #82 consolidation wave.  `evaluate/eval.py` keeps a private `_setup_model` re-export alias so legacy tests keep passing.

- **`model/lightning_module.py::predict_step`** forwards `batch.query` + `batch.duration_days` to the output dict; `subject_id` / `prediction_time` are no longer passed through the batch (removed from `EveryQueryBatch`, `_seeded_getitem`, `collate` — they were only ever used to label predictions, which we now do via `schema_df`).

- **`pyproject.toml`** — `EQ_predict = "every_query.predict.predict:main"`.

- **`tests/test_cli_smoke.py`** — `EQ_predict --help` smoke.
- **`tests/test_predict_cli.py`** — full subprocess integration: happy path + multi-query row-order-preservation sanity check.  Uses `TaskQuerySchema.align(df.to_arrow())` + `pq.write_table` for the test fixtures (no polars round-trip).
- **`tests/training_validity/test_training_validity.py`** — migrated from in-process model forward to subprocess `EQ_predict` + `PredictionSchema` parquet read.  Same unified AUROC criteria, now land on actual CLI output.

## Design decisions

- **Output column naming**: landed on `censor_prob` + `occurs_prob` (matches the model's two-head wiring end-to-end).  #129 tracks the `occurs_prob → label_prob` generalization.
- **Embeddings output**: out of scope for this PR; follow-up `--write-embeddings` if/when the use case materializes.
- **Label pass-through**: `boolean_value` is reconstructed from the dataset's collapsed `(censor, occurs)` and carried through.  Other inherited `LabelSchema` columns (`integer_value`, `float_value`, `categorical_value`) are not pass-through — scope-narrowed since we rebuild the output from `schema_df` instead of the raw input parquets.  A `--carry-columns` opt-in can add them back if the eval-flavoured workflow needs them.
- **Missing-code handling**: warn + silent PAD-encode (doesn't hard-error so callers with broad task sets can still get predictions for the codes the model knows).

## Tmpdir-free input

`task_labels_dir` is now pointed straight at the user's directory — no more rewrite into a tmpdir, no input-side type coercion (trust the user's format; a malformed input surfaces at `PredictionSchema.align(out.to_arrow())` write time).  See [mmcdermott/meds-torch-data#](https://github.com/mmcdermott/meds-torch-data) draft issue on loosening `task_labels_dir` to accept file/dir/glob (not filed yet — sandbox blocked external-repo writes).

## Follow-up

- #130 — `aces_to_eq`'s unused `target_rows` parameter.
- An MTD upstream issue for `task_labels_dir` accepting file/glob inputs (draft in the review thread).

## Test plan

- [x] `pytest --doctest-modules src/every_query/predict/predict.py` — validation + vocab-warn doctests.
- [x] `pytest tests/test_cli_smoke.py` — `EQ_predict --help`.
- [x] `pytest tests/test_predict_cli.py` — E2E + row-order-preservation.
- [x] `pytest -m slow tests/training_validity/` — 7m 10s, unified AUROC criteria pass via `EQ_predict` subprocess path.
- [x] `uv run pytest -q` — 212 passed, 1 deselected locally.
- [x] CI green: code-quality ✓, tests 3.11 ✓, tests 3.12 ✓.

## Depends on / unblocks

- **Depends on**: #96 (`TaskQuerySchema`, merged).
- **Unblocks**: #100 (`EQ_evaluate` consolidation — consumes this PR's output).


mmcdermott:
@copilot review

mmcdermott:
@copilot review

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,198 @@
...
+CONFIGS = str(files("every_query") / "predict" / "configs")
+
+
+def _read_and_validate_tasks(tasks_parquet: Path) -> pl.DataFrame:
+    """Load a task-query parquet and validate against :class:`TaskQuerySchema`.
+
+    Raises with a readable message if the schema doesn't align.
+    """
+    table = pq.read_table(tasks_parquet)
+    TaskQuerySchema.align(table)  # raises with field diff on mismatch
```
mmcdermott: As this returns a table with aligned types, etc., this is a bit less strict than a validate call. That is ok, though, but you could use this too, e.g., `return pl.from_arrow(TaskQuerySchema.align(table))`. I'd say either use validate or return through the align that way you have the benefit of the conversion.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,198 @@
...
+    ckpt_name = cfg.get("ckpt_name")
+
+    logger.info(f"Loading tasks from {tasks_parquet}")
+    tasks_df = _read_and_validate_tasks(tasks_parquet)
+    logger.info(f"Loaded {tasks_df.height} tasks across {tasks_df['query'].n_unique()} query codes")
+
+    train_cfg, model, trainer = _setup_model(model_run_dir, ckpt_name=ckpt_name)
+    _warn_out_of_vocab(tasks_df, train_cfg)
+
+    # Group by (query, duration_days) — the dataset expects a single-query/duration slice.
```
mmcdermott: I'm very confused -- why does the dataset expect a single-query/duration slice? Shouldn't it just make the prediction for whatever task dataframe it is given, like in training? This seems like there is a fundamental design problem.

src/every_query/predict/predict.py
mmcdermott: Done in 059afc6 (one-liner — `pl.from_arrow(TaskQuerySchema.align(pq.read_table(tasks_parquet)))`).  Caller now gets the type-coerced frame so downstream grouping/join keys operate on canonical dtypes — fixes the `duration_days` float64→float32 mismatch you called out.

src/every_query/predict/predict.py
mmcdermott: Fair call — the 'single slice' framing was wrong and the per-group loop was doing nothing the dataset doesn't already handle.

Dropped in 059afc6.  Went with option 1 from the thread (extending `predict_step` to forward `batch.query` + `batch.duration_days`).  In the end the actual labeling uses `D.test_dataset.schema_df` columns directly — same row order (test_dataloader is shuffle=False) but avoids the integer-query round-trip, which would PAD-encode any out-of-vocab codes the caller might pass.  `predict_step`'s new output fields stay as a contract for future batch-level consumers.

Also added a duplicate-key guard in `main()` before the join back — if the input has dup `(subject_id, prediction_time, query, duration_days)` rows, we fail loudly rather than silently multiplying output rows via the many-to-many join.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,198 @@
...
+        )
+
+    predictions = pl.concat(per_group_results, how="vertical_relaxed")
+
+    # Join back to the original tasks_df so inherited label columns carry through.
+    out = tasks_df.join(
+        predictions,
+        on=["subject_id", "prediction_time", "query", "duration_days"],
+        how="left",
+    )
```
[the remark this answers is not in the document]
mmcdermott: Done in 059afc6 — `main()` now validates `(subject_id, prediction_time, query, duration_days)` uniqueness on the input tasks_df up front and raises a clear `ValueError` if any duplicates are present.  Fails cheap instead of silently emitting a malformed predictions parquet.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,198 @@
...
+
+    output_parquet.parent.mkdir(parents=True, exist_ok=True)
+    out.write_parquet(output_parquet)
+    PredictionSchema.align(pq.read_table(output_parquet))  # fail loudly if shape drifted
```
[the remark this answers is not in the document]
mmcdermott: Already done in the previous pass (4bc0905 / 15ce5f7) — see the current `main()`: `aligned = PredictionSchema.align(out.to_arrow()); pl.from_arrow(aligned).write_parquet(output_parquet)`.  The post-write read-back had been a leftover sanity check; it's been removed.  We now emit the canonical-dtype arrow table directly, matching the pattern in `sample_tasks.py:513-521`.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,198 @@
...
+import polars as pl
+import pyarrow.parquet as pq
+import torch
+from hydra.utils import instantiate
+from meds import held_out_split
+from omegaconf import DictConfig
+
+from every_query.data.schema import TaskQuerySchema
+from every_query.evaluate.eval import _setup_model
+from every_query.predict.schema import PredictionSchema
```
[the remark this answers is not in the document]
mmcdermott: Done in 059afc6.  Extracted to `every_query.utils.model_loader.setup_model` (public API).  Predict imports the public name; `evaluate/eval.py` keeps a re-export alias (`from every_query.utils.model_loader import setup_model as _setup_model`) so the legacy tests in `tests/test_eval.py` don't break.  Also updated the mock patch targets in `test_eval.py` to point at the new home.  Gives the #82 consolidation wave a clean path to delete `evaluate/eval.py` without reverb into predict.

src/every_query/predict/predict.py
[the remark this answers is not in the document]
mmcdermott: Done in 059afc6 exactly as suggested: `pl.from_arrow(TaskQuerySchema.align(pq.read_table(tasks_parquet)))`.  The rest of the pipeline now operates on canonical dtypes.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,223 @@
...
+    iteration order.  Identifier columns ``query`` / ``duration_days`` come from the
+    dataset's ``schema_df`` in ``main`` — the integer-encoded ``query`` in the batch
+    is PAD for any out-of-vocab input code, so it isn't round-trip-safe on its own.
+
+    ``logits_to_probs`` does a trailing ``.squeeze()``, so a single-item batch emits a
+    0-d scalar tensor that ``torch.cat`` refuses to stack — ``reshape(-1)`` on every
+    per-batch tensor makes the concat well-defined regardless of batch size.
+    """
+    if not pred_batches:
+        return pl.DataFrame(
```
mmcdermott, line 196: use the schema defined for predictions here, for column names and types.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,223 @@
...
+                "occurs_prob": pl.Float32,
+            }
+        )
+
+    def cat(key: str) -> torch.Tensor:
+        return torch.cat([b[key].reshape(-1) for b in pred_batches])
+
+    return pl.DataFrame(
+        {
+            "subject_id": cat("subject_id").numpy(),
```
mmcdermott: ditto above re column names.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,223 @@
...
+        raise ValueError(
+            f"Input tasks parquet contains {n_dupes} duplicate "
+            f"(subject_id, prediction_time, query, duration_days) rows — each tuple must "
+            f"appear at most once so the join back to labels carries a 1:1 mapping."
+        )
+
+    train_cfg, model, trainer = setup_model(model_run_dir, ckpt_name=ckpt_name)
+    _warn_out_of_vocab(tasks_df, train_cfg)
+
+    with tempfile.TemporaryDirectory() as tmpdir:
```
mmcdermott: I'm very confused -- why do you need a tmpdir here? Why can't you just point directly at the provided file? And can the user not provide a file glob? Is that not supported?

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,223 @@
...
+        TaskQuerySchema.prediction_time_name,
+        TaskQuerySchema.query_name,
+        TaskQuerySchema.duration_days_name,
+    ]
+    n_dupes = tasks_df.height - tasks_df.select(key_cols).unique().height
+    if n_dupes:
+        raise ValueError(
+            f"Input tasks parquet contains {n_dupes} duplicate "
+            f"(subject_id, prediction_time, query, duration_days) rows — each tuple must "
+            f"appear at most once so the join back to labels carries a 1:1 mapping."
```
mmcdermott: What join is happening in this code? There should be none -- it should all be preserved in order throughout, and if not an error should be raised.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,223 @@
...
+    # Horizontal concat in dataset order — ``probs`` comes out of ``trainer.predict``
+    # in ``test_dataloader(shuffle=False)`` order, which matches ``schema_df`` row order.
+    # Drop the duplicate identifiers from ``probs`` before concat.
+    predictions = pl.concat(
+        [identifiers, probs.drop([TaskQuerySchema.subject_id_name, TaskQuerySchema.prediction_time_name])],
+        how="horizontal",
+    )
+
+    # Join back to the original tasks_df so inherited label columns carry through.
+    out = tasks_df.join(
```
mmcdermott: I don't think this should be necessary.

src/every_query/predict/predict.py
mmcdermott: Done in a210fa0 — empty-frame schema now keyed off `PredictionSchema.{censor_prob,occurs_prob}_name` + `pl.Float32` (matches the schema's `pa.float32()`).  Dropped the unused `subject_id` / `prediction_time` fields entirely (they existed for the removed join and duplicated what `tasks_df` already carries).

src/every_query/predict/predict.py
mmcdermott: Done in a210fa0 — same swap on the non-empty frame construction.  `PredictionSchema.{censor_prob,occurs_prob}_name` as keys, explicit `pl.Float32` cast per column so the final `PredictionSchema.align` is dtype-passthrough.  `subject_id`/`prediction_time` are gone from this frame (the removed join was their only consumer).

src/every_query/predict/predict.py
mmcdermott: Honest answer: the tmpdir is there because MTD's `MEDSTorchDataConfig.task_labels_dir` must be a directory (it's `rglob("*.parquet")`'d internally, not `task_labels_fps` pointing at individual files).  So we can't hand it the user's file path directly — only a directory.

What we *can* do:

1. **Point at the user's file's parent**: risky — any stray parquets adjacent to the user's file would get slurped into the dataset.
2. **Symlink** the user's file into a tmpdir: avoids the round-trip write.  Valid, but loses the dtype-canonicalization side-effect that matters for my next bullet.
3. **Rewrite the aligned tasks_df into an empty tmpdir**: isolates the input, and since we just aligned it, we persist canonical dtypes on disk before the dataset reads.  This is what the current code does.

I kept option 3 for now and updated the docstring to spell out the reasoning.

Also simplified per your line 95/109 feedback: dropped the `held_out/` subdir the draft added.  MTD doesn't filter `task_labels_dir` by split — the `held_out` split comes from `Datamodule.test_dataset`, and `task_labels_fps` picks up anything under the dir.  Now writing straight to `{tmpdir}/tasks.parquet`.

If you'd rather change the CLI contract so `tasks_parquet` becomes `tasks_dir`, I'm happy to do that — it'd drop the tmpdir entirely in exchange for a slightly narrower caller API.  Let me know which way you'd prefer.

src/every_query/predict/predict.py
mmcdermott: You're right — dropped the join entirely in a210fa0.  `test_dataloader` is `shuffle=False` and the dataset reads a single shard in sorted file order, so per-row probs come back 1:1 matched with `tasks_df` rows.  Now a plain `tasks_df.hstack(probs)` — no key matching, all inherited `LabelSchema` columns carry through.

Order is *checked*, not trusted: `if probs.height != tasks_df.height: raise RuntimeError(...)`.  The most common cause (task subjects missing from the tensorized `held_out` split) gets named in the error text.

The dup-key guard went with it — without a join there's no many-to-many failure mode.  That simplified `main()` quite a bit.

src/every_query/predict/predict.py
mmcdermott: Agreed — gone in a210fa0.  The join was the reason this guard existed; without it, dup keys just become two hstacked rows (which is fine on this side of the pipeline — if downstream evaluate doesn't like them, that's a downstream concern).  `main()` is ~30 lines shorter now.

src/every_query/predict/predict.py
mmcdermott: Can you file an upstream issue in MTD about this? This contract should be more flexible -- the task labels input dir should take a glob or be able to support both directory and file inputs. For us for now, the better option is option 4 -- have the user pass a directory instead of a file, with an indicative name. Then, if it isn't a directory, or if there things in it that aren't parquet files, through an error, and log an info message or warning if there are more parquet files in it. Then we don't need to do this copying. Honestly, we also don't need to do the type coercion too -- the contract is the user gives us something in the right format, and the code upstream should be flexible enough to handle it, right? 

Another alternative is we could change our dataset subclass to address this here, but I don't think that's the right call.

mmcdermott: Agreed — this is a better design.  Implementing option 4:

- `predict.yaml` config: `tasks_parquet` → `tasks_dir`
- `main` validates: must be a directory, must contain only `.parquet` files, warn if >1 parquet present
- Pass `tasks_dir` straight to `task_labels_dir` — no tmpdir, no rewrite, no symlink
- Drop the input-side `TaskQuerySchema.align` coercion — if the user gives us something malformed, MTD will surface that; we keep `PredictionSchema.align` at write so the *output* is always canonical
- Still read the input parquets ourselves for the `hstack` (just `pl.concat([pl.read_parquet(fp) for fp in sorted(tasks_dir.rglob('*.parquet'))])`) — the dataset's grouping logic in `get_task_seq_bounds_and_labels` might collapse multi-query-per-`(subject,time)` rows, so we can't pull tasks from `schema_df` directly

On the upstream MTD issue: I tried to file one but my sandbox blocks writes to non-primary repos.  Content is ready — happy to either (a) hand you a draft to paste, or (b) get you to grant the Bash permission rule so I can file it.  Draft follows (summary of the ask):

> **Title**: `task_labels_dir` should accept a file or glob, not require a directory
>
> `MEDSTorchDataConfig.task_labels_dir` currently requires a directory — `task_labels_fps` does `path.rglob("*.parquet")` and `__post_init__` validates `is_dir()`.  Callers with a single flat task parquet have to either rewrite/copy it into an empty tmpdir or point at the file's parent (risky — stray parquets get slurped).
>
> Proposal: loosen `task_labels_dir` to accept a file, a dir, or a glob.  Resolution: file → `[path]`; dir → current rglob; glob → `Path.glob`.

Pushing the code change in a moment.  Let me know on the issue-filing preference.

mmcdermott: Yes — great call.  Done in c1b8f02.

`main()` no longer re-reads the input parquets.  MTD's `get_task_seq_bounds_and_labels` [docstring](https://github.com/mmcdermott/meds-torch-data/blob/main/src/meds_torchdata/pytorch_dataset.py) explicitly guarantees `schema_df` preserves the input labels frame's length + order (`"This function is guaranteed to output an index of the same order and length as label_df."`), so the dataset already has everything the output needs.

New helper `_identifiers_from_schema_df(schema_df)` picks the four `TaskQuerySchema` key columns + reconstructs the original nullable `boolean_value` from the collapsed `(censor, occurs)` pair (inverse of the collapse in `EveryQueryPytorchDataset.__init__`).  Then `main` is just `identifiers.hstack(probs)` → `PredictionSchema.align` → write.

To catch a future MTD bump that silently starts reordering (your "slash catch errors in that assumption" — also a fine call), added `test_eq_predict_preserves_input_row_order`: a multi-query / multi-duration case with deliberately non-alphabetical input rows — `[(TEMP, 60), (HR, 30), (TEMP, 30)]` — asserting the output column values match the input order position-by-position.  If MTD ever sorts `schema_df` internally, that test fails loudly.

**One scope note to confirm**: since we're no longer reading the raw parquet, inherited `LabelSchema` columns beyond `boolean_value` (`integer_value` / `float_value` / `categorical_value`) won't pass through to the output anymore.  The typical inference use case doesn't need those, and I documented the scope narrowing in the module docstring.  If you want them pass-through for an eval-flavoured workflow, an opt-in `--carry-columns` that re-reads the parquets is a clean follow-up — let me know if you want that in this PR or separately.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,223 @@
...
+from every_query.predict.schema import PredictionSchema
+from every_query.utils.model_loader import setup_model
+
+logger = logging.getLogger(__name__)
+logging.basicConfig(level=logging.INFO)
+
+CONFIGS = str(files("every_query") / "predict" / "configs")
+
+
+def _validate_tasks_dir(tasks_dir: Path) -> None:
```
mmcdermott, line 57: you could have doctests here to show validation errors and checks, to minimize test burden elsewhere, but it is fine for now.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,223 @@
...
+        raise ValueError(f"tasks_dir {tasks_dir} contains no parquet files")
+    if len(parquets) > 1:
+        logger.warning(
+            f"tasks_dir {tasks_dir} contains {len(parquets)} parquet files; all will be "
+            f"concatenated.  The typical inference use case is a single flat parquet."
+        )
+
+
+def _warn_out_of_vocab(task_codes: set[str], train_cfg: DictConfig) -> None:
+    """Warn if any task-query codes are missing from the trained model's vocabulary.
```
mmcdermott, line 135: ditto above.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,223 @@
...
+
+def _identifiers_from_schema_df(schema_df: pl.DataFrame) -> pl.DataFrame:
+    """Build an output identifier frame from the dataset's post-``__init__`` schema_df.
+
+    Reconstructs the original nullable ``boolean_value`` from the collapsed
+    ``(censor, occurs)`` pair (inverse of the collapse in
+    ``EveryQueryPytorchDataset.__init__``) so the output matches the
+    ``TaskQuerySchema`` ``null``/``True``/``False`` convention.
+
+    ``prediction_time`` is cast back to ``Datetime("us")`` — the dataset stores
```
mmcdermott: Why does this happen? This seems like something is wrong.

tests/test_predict_cli.py
```diff
@@ -0,0 +1,191 @@
...
+            "prediction_time": _PRED_TIME,
+            "query": code,
+            "duration_days": _DURATION_DAYS,
+            "boolean_value": None,
+        }
+        for code in _QUERY_CODES
+    ]
+    df = pl.DataFrame(rows).cast(
+        {
+            "subject_id": pl.Int64,
```
mmcdermott: use the schema. Here you can use it via the align write pyarrow path directly.

tests/test_predict_cli.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Guards the implementation's load-bearing assumption that ``D.test_dataset.schema_df``
+    preserves the input labels frame's row order (per MTD's
+    ``get_task_seq_bounds_and_labels`` guarantee).  If a future MTD bump reorders
+    schema_df rows, the output ``(query, duration_days)`` column would desync from the
+    input parquet and this test would fail.
+    """
+    # Non-alphabetical query order, mixed durations — exercises the order check past
+    # the alphabetical HR/TEMP happy path of the main integration test.
+    rows = [
```
mmcdermott: with something like this it is IMO cleaner to use the dictionary form (e.g., column_name: [val1, val2,...]) instead of the list of dicts form.

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,223 @@
...
+            f"``test_dataloader`` should have yielded one prediction per schema_df row."
+        )
+
+    out = identifiers.hstack(probs)
+
+    output_parquet.parent.mkdir(parents=True, exist_ok=True)
+    # Final canonicalization pass — ``align`` casts / reorders columns to match the
+    # schema's canonical arrow layout, so the written parquet is schema-conformant
+    # regardless of whatever dtypes the input happened to have.
+    aligned = PredictionSchema.align(out.to_arrow())
```
mmcdermott, line 309: you can just use pq directly to write, you don't need to go back to polars.

src/every_query/predict/predict.py
mmcdermott: Great question — it doesn't, now.

The batch carried `subject_id` + `prediction_time` only so `predict_step` could emit them in the output dict, so the old `_run_predict` / draft `EQ_predict` could label predictions.  With our new flow using `D.test_dataset.schema_df` directly for the output identifiers, the batch-level copies are dead weight.

Ripping them out in the next push:
- `EveryQueryBatch`: drop the `subject_id` / `prediction_time` fields + shape validations.
- `EveryQueryPytorchDataset._seeded_getitem`: drop the `held_out`-specific injection.
- `EveryQueryPytorchDataset.collate`: drop the `held_out`-specific tensor conversion (which is what the previous `schema_df` int-conversion was feeding).  That conversion itself also goes — `schema_df` stays as `Datetime("us")` end-to-end.
- `predict_step`: drop them from the output dict.
- Doctest / `test_lightning_logic.py`: update accordingly.

The model itself never used them (loss uses `censor` / `occurs`; logging uses nothing else), so this is a pure simplification with no functional change.

src/every_query/predict/predict.py
mmcdermott: Done in c6018ec — added doctests covering happy path, non-dir, empty dir, non-parquet file, and the multi-parquet warn.  That pins the contract without needing unit tests elsewhere.

src/every_query/predict/predict.py
mmcdermott: Done in c6018ec — doctests for happy path (no warn), out-of-vocab warn, and missing-metadata warn.

tests/test_predict_cli.py
mmcdermott: Done in c6018ec — new `_write_tasks_parquet(fp, columns)` helper routes through `TaskQuerySchema.align(df.to_arrow())` for dtypes, so the test sites no longer restate per-column `pl.Int64` / `pl.Float32` casts.  Both test fixtures now use it.

tests/test_predict_cli.py
mmcdermott: Done in c6018ec — both fixtures switched from list-of-dicts to column-dict form (`{column_name: [val1, val2, ...]}`).  Each column's values line up at a glance now.

tests/test_predict_cli.py
mmcdermott: Sorry — I read "use the schema" and missed the "pyarrow path directly" half.  Fixed in 3e7e182.

Was: `pl.from_arrow(TaskQuerySchema.align(df.to_arrow())).write_parquet(fp)`
Now: `pq.write_table(TaskQuerySchema.align(pl.DataFrame(columns).to_arrow()), fp)`

Applied the same fix to `predict.py`'s output write path too — the polars intermediary was actually slightly lossy (polars re-infers dtypes from the aligned data, which defeats the point of aligning in the first place).  Writing the arrow table directly lands the align()-produced dtypes on disk unchanged.

PR #110, opened 2026-04-18
test(e2e): foundation — conftest CLI-chain fixtures + smoke test

mmcdermott:
## Summary

First PR in the #104 E2E integration testing effort.  Adds the conftest fixture DAG that downstream per-stage tests (#106, #107, #108, #109) will chain off.  Ships a minimal smoke test that exercises the whole chain.

Closes #105.  Refs #104.

### `conftest.py` — three new session-scoped CLI-chain fixtures

```
simple_static_MEDS ──► eq_preprocessed_dataset  (EQ_process_data do_demo=True)
                             │
                             ├────► eq_sampled_tasks_dir   (EQ_generate_tasks × {train, tuning})
                             │              │
                             │              ▼
                             └────► eq_trained_model_dir   (EQ_train --config-name=_demo_train)
```

Each fixture runs the real `EQ_*` console script in a subprocess via a shared `run_and_check(cmd, env, timeout)` helper (raises with captured stdout/stderr on nonzero exit so CI failures point at the concrete error).  All fixtures are `scope="session"` — each stage runs at most once per pytest invocation.

The in-process model/data fixtures (`demo_model`, `sample_batch`, `tensorized_cohort_dir`) are unchanged — narrower unit tests that just need a model forward pass should keep using those; the new CLI-chain fixtures are for integration tests that want to exercise the actual CLI surface.

### `tests/test_e2e_foundation.py` — minimal smoke test

Three tests, one per stage, each asserting the expected output directory was produced.  Runtime: **24s on CPU**.  Deliberately minimal — the per-stage PRs (#106–#109) will add real differential + schema assertions on top of the same fixtures.

## Quirks worth noting

**`_artifacts/` dir leaks into `task_labels_dir`.**  `EQ_generate_tasks` writes both its labeled parquets *and* its caching artifacts into the same `out_dir`.  MTD's `task_labels_fps` globs recursively, so the 4-column `_artifacts/index/` files would be picked up alongside the 6-column labeled rows and break `pl.concat`.  The `eq_sampled_tasks_dir` fixture works around this by staging just `{split}/*.parquet` into a clean directory after running the sampler.  Worth fixing upstream — maybe move `_artifacts/` outside `out_dir`, or give MTD a way to restrict its glob.

[...]

Issue #104, opened 2026-04-18
E2E integration testing: real subprocess-driven pipeline tests (replaces the closed #68)


mmcdermott:
## Motivation

#68 (CLI wheel-install smoke) + PR #98 were closed as out-of-scope.  What we actually need is end-to-end tests that exercise the **real pipeline**, not just `--help` exits.  The refactor umbrella (#54) changes CLI contracts, data-flow shapes, and submodule boundaries — without integration tests we are flying blind on whether any given refactor PR breaks the pipeline.  This should be a **prerequisite for, or run alongside,** the remaining Phase 2 + Phase 3+ work.

## Model

Follow the MEDS_EIC_AR (MEICAR) pattern.  MEICAR has per-CLI integration tests wired through `conftest.py`:

- `tests/test_process_data.py` — runs `MEICAR_process_data` on `simple_static_MEDS`, asserts outputs exist.
- `tests/test_pretrain.py` — runs `MEICAR_pretrain` with `--config-name=_demo_pretrain`, asserts checkpoints exist, asserts resume works.
- `tests/test_generate_trajectories.py` — runs `MEICAR_generate_trajectories`, loads output parquet, asserts row counts + cross-sample equality + **differential** assertion (`max_rolling_rows > max_single_rows` — proves the rolling knob was actually honored, not silently dropped).

Chaining is via **pytest fixture dependency**: `pretrained_model(preprocessed_dataset)` — pytest wires stage B's input to stage A's output.  Everything is Hydra CLI subprocess invocation with `--config-name=_demo_<stage>` stubs shipped inside the package.

Reference files in MEICAR:
- [`conftest.py`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/conftest.py) — `run_and_check` helper + session-scoped chained fixtures
- [`tests/test_process_data.py`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/tests/test_process_data.py)
- [`tests/test_pretrain.py`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/tests/test_pretrain.py)
- [`tests/test_generate_trajectories.py`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/tests/test_generate_trajectories.py)

[...]

PR #111, opened 2026-04-18
test(e2e): real EQ_process_data integration test + differential

### LINKED ISSUE #106: E2E tests: test_process_data.py — real EQ_process_data subprocess test (refs #104)
## Scope

Second child of #104.  First real CLI-chain test: runs `EQ_process_data` via subprocess on `simple_static_MEDS`, asserts the output shape.

## What lands

- `tests/test_process_data.py` with:
  - A test that consumes the `preprocessed_dataset` fixture (from the foundation PR) and asserts the expected output files exist: tensorized shards, schema file, split files.
  - A **differential** assertion — re-run with a different `min_events` threshold, assert row count / shard count differs correctly.  This is the "real integration" part; shape-only checks would pass even if flags were silently dropped.
- Doctest-compatible output if useful.

## Depends on

- #105 (foundation).

## Acceptance

- New test passes on a clean `uv run pytest tests/test_process_data.py`.
- Runtime under 2 minutes on CPU.
- Failure messages identify which stage / which assertion failed.

mmcdermott:
## Summary

Second PR in the #104 E2E testing effort.  Stacks on #110 (foundation).  Adds real assertions for the `EQ_process_data` CLI on top of the `eq_preprocessed_dataset` fixture.

Closes #106.  Refs #104.

## What lands

`tests/test_process_data.py` with 4 tests:

1. **`test_expected_stage_dirs_exist`** — verifies the tensorization stage output layout (metadata/, tokenization/{event_seqs,schemas}/, normalization/, fit_normalization/, fit_vocabulary_indices/).  Catches structural regressions that would otherwise surface only at dataloader-time.

2. **`test_codes_metadata_schema`** — loads codes.parquet, asserts required aggregation columns are present and `code/vocab_index` is dense 1..N (index 0 reserved for padding).  Regressions to the token-id contract would otherwise be silent.

3. **`test_per_split_tokenization_outputs`** — each MEDS split (train / tuning / held_out) has tokenized event_seqs + schema shards with subject_id + code columns.

4. **Differential — `test_min_subjects_filter_is_honored`**.  Re-runs `EQ_process_data` with `MIN_SUBJECTS_PER_CODE=5` via env override and asserts the resulting vocabulary is strictly smaller than the demo run's (MIN=2).  Catches silent flag-drop regressions in the pipeline config — shape-only tests would pass even if the filter were quietly ignored.

## Test plan

- [x] `uv run pytest tests/test_process_data.py -v` — 4 passed in ~30s
- [ ] CI green on push

## Merge order

Rebase onto `dev` after #110 merges.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,6 +1,6 @@
 """Root conftest — session-scoped fixture DAG for the EveryQuery test suite.
 
-Fixture dependency graph (all session-scoped)::
+In-process fixtures (objects live in the test interpreter)::
 
     demo_model ──────────► demo_lightning_module
 
@@ -11,8 +11,23 @@ Fixture dependency graph (all session-scoped)::
                                     │
                                     ▼
                                sample_batch
+
+CLI-chain fixtures (run real ``EQ_*`` console scripts via subprocess; each returns the
+output directory so downstream tests / fixtures can chain off it)::
+
+    simple_static_MEDS ──► eq_preprocessed_dataset
+                                    │
+                                    ├────► eq_sampled_tasks_dir
+                                    │             │
+                                    │             ▼
+                                    └────► eq_trained_model_dir
+
+All fixtures in this file are ``scope="session"`` — each runs at most once per pytest
+invocation.
 """
 
+import os
+import shutil
 import subprocess
 import sys
 import tempfile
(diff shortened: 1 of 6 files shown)


mmcdermott:
@copilot review

Issue #105, opened 2026-04-18
E2E tests: foundation — conftest fixture DAG + _demo_*.yaml configs (refs #104)


mmcdermott:
## Scope

First child of #104.  Ship the plumbing before any individual test file.

## What lands

- A `run_and_check(cmd)` helper in `conftest.py` — subprocess wrapper that raises with captured stdout/stderr on nonzero exit (MEICAR pattern).
- Session-scoped chained fixtures returning `Path` outputs:
  - `preprocessed_dataset(simple_static_MEDS) -> Path` — runs `EQ_process_data` (falls back to existing `tensorized_MEDS_dataset` if EQ_process_data isn't usable yet).
  - `generated_tasks_dir(preprocessed_dataset) -> Path` — runs `EQ_generate_tasks`.
  - `pretrained_model_dir(preprocessed_dataset, generated_tasks_dir) -> Path` — runs `EQ_train --config-name=_demo_train`.
  - Follow-on stages (`predictions_dir`, `evaluation_metrics_dir`) filed as separate children — foundation does not block on them.
- `_demo_<stage>.yaml` configs inside the package (`src/every_query/<stage>/configs/_demo_<stage>.yaml`) for CPU-fast testing:
  - tiny `hidden_size`, `num_hidden_layers`, `max_seq_len`
  - 1-2 train steps max
  - point at the tests-scope default seed so runs are reproducible

## Why this comes first

Every per-stage test needs its upstream fixture to exist.  Filing this as a distinct PR lets reviewers assess the plumbing separate from per-CLI assertion logic.

## Out of scope

- Any specific per-stage test file (that's #106 / #107 / etc.).
- CI config changes — the new tests should just run under the existing pytest invocation.
- `EQ_predict` / `EQ_evaluate` fixtures — those are gated on #99/#100 landing.

## Acceptance

- New fixtures are importable and chain cleanly.
- `pytest tests/` still passes.
- Demo configs documented in their submodule READMEs.

mmcdermott:
Closing — addressed by PR #110, merged to `dev`.

PR #115, opened 2026-04-18
fix(generate_tasks): move sampler caching artifacts out of out_dir

### LINKED ISSUE #114: EQ_generate_tasks caching artifacts leak into task_labels_dir and break EQ_train
## Symptom

Anyone following the documented CLI workflow — `EQ_generate_tasks out_dir=$TASK_DIR` then `EQ_train datamodule.config.task_labels_dir=$TASK_DIR` — hits this at training time:

```
polars.exceptions.ShapeError: unable to append to a DataFrame of width 4 with a DataFrame of width 6
```

inside `EveryQueryPytorchDataset.labels_df`'s `pl.concat` call.  Surfaces as Hydra job failure during `trainer.fit` setup.

## Root cause

`sample_tasks._artifact_paths` writes the sampler's caching artifacts (sampled tasks JSON, **unlabeled index parquet**, run-meta JSON) underneath `out_dir`:

```
out_dir/
├── train/0__0000.parquet           ← 6-column labels (the real output)
├── tuning/0__0000.parquet          ← 6-column labels
└── _artifacts/
    ├── tasks/train/0__0000.json
    ├── index/train/0__0000.parquet ← 4-column unlabeled index (THE LEAK)
    └── runs/train/0__0000.json
```

MEDS-torch-data's [`MEDSTorchDataConfig.task_labels_fps`](https://github.com/kamilest/meds-torch-data/blob/main/src/meds_torchdata/config.py) globs recursively via `rglob("*.parquet")`.  It picks up the 4-column `_artifacts/index/*.parquet` alongside the 6-column label parquets.  `pl.concat(..., how="vertical")` then fails on the schema mismatch.

## Fix

Move artifacts out of `out_dir`: put them in a sibling `{out_dir}.cache/` directory instead of `{out_dir}/_artifacts/`.  `out_dir` then contains only label parquets, which is the contract downstream consumers assume.

PR: #114.

## Why not caught earlier

The unit tests in `tests/test_sample_tasks.py` checked `_artifact_paths`' output directly — they didn't exercise the downstream MTD glob contract.  The E2E integration tests being added under #104 catch this directly: a regression test in `tests/test_e2e_foundation.py` asserts every parquet under the sampler's `out_dir` conforms to the label schema.  Had this test existed before the sampler refactor that introduced the `_artifacts/` layout, this bug would not have shipped.

mmcdermott:
## Summary

Fixes a production-breaking bug in `EQ_generate_tasks`: the sampler's caching artifacts leaked into the user's `out_dir`, which downstream `EQ_train` / `EQ_evaluate` consume as `task_labels_dir`, breaking training with a confusing `polars ShapeError` at `trainer.fit` time.

Closes #114.

## What changed

`sample_tasks._artifact_paths` now writes caching artifacts to a sibling `{out_dir}.cache/` directory instead of `{out_dir}/_artifacts/`:

```
# before
out_dir/
├── train/0__0000.parquet           ← 6-column labels
└── _artifacts/
    ├── index/train/0__0000.parquet ← 4-column index (leaked into MTD glob)
    ├── tasks/train/0__0000.json
    └── runs/train/0__0000.json

# after  
out_dir/
└── train/0__0000.parquet           ← 6-column labels only
out_dir.cache/                      ← sibling
├── index/train/0__0000.parquet
├── tasks/train/0__0000.json
└── runs/train/0__0000.json
```

MTD's recursive `task_labels_fps` glob now picks up only real labels.  Sampler CLI signature and user-facing contract unchanged — users still pass `out_dir=$TASK_DIR` and then `task_labels_dir=$TASK_DIR`; the cache just appears as `$TASK_DIR.cache/` alongside.

## What's not in this PR (deliberately)

The E2E integration-test foundation (#105, PR to follow) adds a regression test (`test_generate_tasks_out_dir_contains_only_label_parquets`) that asserts every parquet under the sampler's `out_dir` conforms to the label schema.  That test belongs with the integration-test PR (its real goal is to run the full CLI chain); the fix itself is a minimal source change that stands alone.

## Test plan

- [x] `uv run pytest tests/test_sample_tasks.py` — 40 passed (path-references updated)
- [x] `uv run pytest tests/` — full suite green
- [ ] CI green on push

A regression test that would have caught this lives in the follow-on PR #110 (E2E foundation) — once both merge, the integration-test suite will fail-fast on any future reintroduction.

### DIFF
--- a/src/every_query/generate_tasks/configs/sample_tasks_config.yaml
+++ b/src/every_query/generate_tasks/configs/sample_tasks_config.yaml
@@ -8,7 +8,10 @@
 # (see `.env.example`, `tasks.py`, and `_env.py`).  Fallbacks:
 #   data_dir  -> $INTERMEDIATE   (event shards live under `{data_dir}/data/{split}/*.parquet`)
 #   codes_dir -> $PROCESSED      (query universe lives under `{codes_dir}/metadata/codes.parquet`)
-#   out_dir   -> $TASK_DIR       (labeled shards and _artifacts/ land here)
+#   out_dir   -> $TASK_DIR       (labeled shards land here as {split}/*.parquet;
+#                                 caching artifacts live in a sibling {out_dir}.cache/
+#                                 dir so MTD's recursive task_labels_fps glob doesn't
+#                                 pick up the non-label parquets)
 # Tests can inject values directly with `data_dir=/path out_dir=/path codes_dir=/path`.
 
 data_dir: null
--- a/src/every_query/generate_tasks/sample_tasks.py
+++ b/src/every_query/generate_tasks/sample_tasks.py
@@ -455,12 +455,35 @@ def _artifact_paths(
     input_shard: str,
     task_shard: int,
 ) -> tuple[Path, Path, Path, Path]:
-    """Resolve the ``(tasks_fp, index_fp, labels_fp, run_meta_fp)`` quadruple for one worker."""
+    """Resolve the ``(tasks_fp, index_fp, labels_fp, run_meta_fp)`` quadruple for one worker.
+
+    Only the *labels* parquet lives under ``out_dir`` — the caching artifacts (sampled tasks
+    JSON, unlabeled-index parquet, run-meta JSON) live in a sibling ``{out_dir}.cache/``
+    directory.  This separation is load-bearing: downstream tooling (MTD's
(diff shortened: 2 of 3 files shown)


mmcdermott:
@copilot review

mmcdermott:
Superseding with a smaller PR that removes the caching entirely — rerunning is cheap anyway and seed derivation already pins determinism.  Moving this path out of out_dir is the wrong fix; the right fix is to not write these files at all.  See discussion in chat.

PR #126, opened 2026-04-19
fix(evaluate): require model_run_dirs override instead of shipping stale dated paths

### LINKED ISSUE #66: `eval_config.yaml` hardcodes stale dated `model_run_dirs` — eval can't run without editing the shipped config
## What I found

`src/every_query/eval_suite/conf/eval_config.yaml` ships with two dated, experiment-specific paths as the default:

```yaml
model_run_dirs:
  - \${oc.env:OUTPUT_DIR}/outputs/2026-04-02/23-43-54 # 256, 22
  - \${oc.env:OUTPUT_DIR}/outputs/2026-04-02/23-54-43 # 1028, 22
  # - \${oc.env:OUTPUT_DIR}/outputs/2026-04-08/16-38-46/
```

These are timestamped training run directories from a specific experiment. A new user (or even the same user after a new training run) cannot \`EQ_evaluate\` without first editing this file. There's no \`???\` marker to signal \"user must override\" and no comment explaining it.

## Proposed change

Replace with an explicit required-override:

```yaml
model_run_dirs: ???  # list of training run directories (e.g. \${oc.env:OUTPUT_DIR}/outputs/YYYY-MM-DD/HH-MM-SS/)
```

Users then pass `model_run_dirs=[\"\$OUTPUT_DIR/outputs/.../\"]` on the CLI. This surfaces the requirement at the Hydra-error layer instead of hiding it behind a stale-path \"FileNotFoundError\".

## Follow-ups

- Same pattern may apply to `select_model_config.yaml` — it also has two hardcoded dated paths at lines 1-3 (need to verify).
- README / docs (#59) should show an example `EQ_evaluate` invocation with a real override.

cc @mmcdermott (last editor on eval_config.yaml via d607a1f)

mmcdermott:
## Summary

Closes #66.

Both `eval_config.yaml` and `select_model_config.yaml` shipped with two dated experiment run directories (`${oc.env:OUTPUT_DIR}/outputs/2026-04-02/...`) as the `model_run_dirs` default.  A fresh user — or the same user after a new training run — couldn't `EQ_evaluate` / `EQ_select_model` without first editing those files, and nothing signaled that an override was mandatory; the failure mode was a silent `FileNotFoundError` when the first dated path didn't exist on the current machine.

This swaps the hardcoded paths for `???` (Hydra's required-value sentinel) so the requirement surfaces at config-resolution time with a clear "mandatory value missing" error, and adds an inline comment showing the override pattern.

## Files

- `src/every_query/evaluate/conf/eval_config.yaml` — `model_run_dirs: ???`
- `src/every_query/evaluate/conf/select_model_config.yaml` — same

## Test plan

- [x] `uv run pytest tests/test_cli_smoke.py -v` — 7/7 passed.  Hydra's `--help` renders help text without resolving required values, so `EQ_evaluate --help` and `EQ_select_model --help` still exit 0.
- [ ] CI green on push.

### DIFF
--- a/src/every_query/evaluate/conf/eval_config.yaml
+++ b/src/every_query/evaluate/conf/eval_config.yaml
@@ -2,20 +2,10 @@ defaults:
   - _self_
   - eval_codes: 10000_ID__8db2be6fadf8
 
-# Model run directory (or list of directories).
-# Paths are rooted at $OUTPUT_DIR so they move with .env across machines
-model_run_dirs:
-  - ${oc.env:OUTPUT_DIR}/outputs/2026-04-02/23-43-54 # 256, 22
-  - ${oc.env:OUTPUT_DIR}/outputs/2026-04-02/23-54-43 # 1028, 22
-  # - ${oc.env:OUTPUT_DIR}/outputs/2026-04-08/16-38-46/
-  # - ${oc.env:OUTPUT_DIR}/outputs/2026-04-03/00-29-53 #2048,6
-  # - ${oc.env:OUTPUT_DIR}/outputs/2026-04-02/23-56-55 # 1028, 6
-  # - ${oc.env:OUTPUT_DIR}/outputs/2026-04-02/23-58-42 # 1028, 12
-  # - ${oc.env:OUTPUT_DIR}/outputs/2026-04-04/14-08-24 # 1028, 16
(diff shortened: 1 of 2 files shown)


mmcdermott:
merging and closing this now, but we really should not have the evaluation code take in multiple models. The fact that this takes in a list is not the right paradigm here.

Issue #129, opened 2026-04-19
Consider renaming PredictionSchema.occurs_prob → label_prob when generalizing to non-occurrence tasks


mmcdermott:
Post-refactor / post-NeurIPS follow-up.

## Context

The `PredictionSchema` introduced alongside `EQ_predict` (#81 / #96 stack) names the non-censor head's output `occurs_prob` — matching the current two-head model architecture (`censor_head` + `occurs_head`) and the existing `EveryQueryBatch.occurs` tensor.  This is the right call *today* because every downstream metric and debugging tool walks the wiring `occurs_head → batch.occurs → eval's occurs_prob` with a single term.

## Proposal

Once EveryQuery supports task types beyond "does this code occur within d days" — e.g., regression tasks (length-of-stay prediction), multi-class (next-most-likely code), survival-style tasks — the `occurs_prob` name will feel too narrow.  At that point, rename the column to `label_prob` (or similar general term) alongside the schema / model generalization.

## Scope notes

- **Not this PR**: rolling a generalization into the Phase 2.2 `EQ_predict` scope would gate the inference CLI on a schema design we don't have yet.  Better to ship the concrete occurrence case first, prove the pipeline end-to-end, then generalize with evidence.
- **When to revisit**: when the first non-occurrence task lands (or is being designed).  If that's driven by the NeurIPS submission, consider before the submission freeze; otherwise defer until the next architecture iteration.
- **Scope if revisited**: rename `PredictionSchema.occurs_prob` → `label_prob`; rename `EveryQueryBatch.occurs` → `EveryQueryBatch.label` (or keep `occurs` for occurrence-specific tasks and add `label` as the generalized axis); update all downstream metric code; update docs + tests.

## Related

- #54 — refactor umbrella
- #81 — EQ_predict (will introduce `PredictionSchema`)
- #96 — TaskQuerySchema (the input counterpart)
- #83 — EQ_evaluate consolidation (first downstream consumer)

PR #128, opened 2026-04-19
docs(readme): reconcile with current refactor progress

mmcdermott:
## Summary

Docs-only update to the top-level `README.md`.  Reconciles the pipeline description, env-var table, and roadmap with the current state of `dev` after the recent merges (#112, #113, #119, #124, #126, #127) and documents what's covered on each CLI endpoint vs. what's planned.

## What changed

- **Console scripts table**: added a "Tests" column so the coverage that lands with each CLI on `dev` today (smoke / unit / E2E subprocess / slow signal test) is visible without grepping, with issue + PR refs for each.
- **Pipeline section**: kept the current diagram, added a "Planned" diagram for the Phase 2 target (preprocess → generate_tasks → train → predict → evaluate).  Inline notes on the current diagram call out the `code` column / `TaskQuerySchema` alignment (#96), the #124 seed-ordering fix, and the #126 `model_run_dirs` required-override.
- **Environment variables**: dropped `PROCESSED` and `INTERMEDIATE` rows (no longer in `REQUIRED_ENV_VARS` after #127); added a note pointing at the remaining phases of #117.
- **Development**: added a full "Test layout" subsection listing every file under `tests/` with its coverage scope, and clarified the `pytest -m 'slow or not slow'` opt-in for the training-validity test.
- **Roadmap**: restructured into three focused tables — Phase 2 status (#80/#81/#82/#83), E2E testing status (#104 umbrella), and hygiene follow-ups.  Merged items are marked ✅; in-flight items link to their issue or draft PR.  Architecture research (RoPE, ModernBERT alternatives) moved to its own subsection so it doesn't compete visually with the refactor tracker.

## Test plan

- [x] `uv run pytest tests/test_cli_smoke.py` — 7/7 pass (docs change doesn't touch code).
- [x] `pre-commit run --all-files` via the commit hook — mdformat applied, commit landed.
- [ ] CI green on push.

Docs-only — no code changes.  No Python files modified; no tests altered.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #127, opened 2026-04-19
fix(env): drop PROCESSED and INTERMEDIATE from REQUIRED_ENV_VARS

mmcdermott:
## Summary

Refs #117 (phase 1 of the env-vars audit — the "immediate, one-line, zero-risk" item).

`grep '${oc.env:PROCESSED}' / '${oc.env:INTERMEDIATE}'` across all Hydra configs returns zero hits.  Their only consumer is a dotenv fallback inside `generate_tasks/sample_tasks.py::_resolve_path`, which already tolerates missing env vars when config values are supplied (the normal CLI path).  Keeping them in `REQUIRED_ENV_VARS` was pure friction — demo fixtures and E2E integration tests had to set throwaway `/tmp` placeholders in `ENSURE_ENV_PLACEHOLDERS` / `test_train_cli.py`'s subprocess env just to satisfy a gate that didn't correspond to any real Hydra resolution, and `EQ_* --help` on a fresh machine failed on these vars too.

## Changes

- `src/every_query/utils/_env.py` — drop `PROCESSED`, `INTERMEDIATE` from `REQUIRED_ENV_VARS`.  Comment left in place explaining why.
- `conftest.py` — drop both from `ENSURE_ENV_PLACEHOLDERS`.
- `tests/test_train_cli.py` — drop both from the per-test subprocess env setdefault loop.

## Test plan

- [x] `uv run pytest tests/test_train_cli.py tests/test_e2e_foundation.py tests/test_cli_smoke.py tests/test_training.py -v` — 37/37 passed in 78s locally.
- [ ] CI green on push.

## Follow-ups tracked in #117

- Phase 2: drop `PROJECT_DIR` after #100 lands (its only remaining consumer is the eval-suite code being retired there).
- Phase 3: convert remaining `${oc.env:VAR}` interpolations to `${oc.env:VAR,???}` / `${oc.env:VAR,default}` form so Hydra owns the "required" enforcement directly.
- Phase 4: retire `ensure_env()` entirely.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #131, opened 2026-04-20
Release candidate — refactor Phase 1 + 2.1/2.2 + partial 2.4 (roll-up)

### LINKED ISSUE #31: `train.py`: `do_overwrite=True` deletes the config before saving the new one, losing reproducibility artifacts

### LINKED ISSUE #32: `eval.py`: `code in cfg.ood_codes` crashes with TypeError when `ood_codes` is None

### LINKED ISSUE #52: Move derive_seed to utils
## Context

`derive_seed` currently lives in [`src/every_query/sample_tasks.py` (L59)](https://github.com/payalchandak/EveryQuery/blob/feat/sample-task-generator/src/every_query/sample_tasks.py#L59). It is a general-purpose deterministic hashing utility (blake2b-based, cross-process stable) that is not specific to task sampling.

## Proposal

Move `derive_seed` into `src/every_query/utils/` so that:
- Other modules that need stable seeding (e.g., context sampling, data splitting) can import it without a misleading dependency on `sample_tasks`.
- `sample_tasks.py` stays focused on task sampling logic.

Update the import in `sample_tasks.py` and `tests/test_sample_tasks.py` accordingly.

### LINKED ISSUE #53: Refactor: remove dead tasks.py code superseded by sample_tasks.py

### LINKED ISSUE #62: Orphaned modules: `aces_to_eq/` and `process_composite/` have no inbound imports
## What I found

While doing the doc overhaul for #59, I went looking for who imports each module under `src/every_query/`. These two have **zero** inbound Python imports — neither from the main codebase, tests, notebooks, nor from any of the 7 `EQ_*` console scripts registered in #61:

- `src/every_query/aces_to_eq/aces_to_eq.py` (+ `config.yaml`)
- `src/every_query/process_composite/process_composite.py`, `get_per_code_from_composite.py` (+ `process_composite_config.yaml`)

Both are `@hydra.main`-decorated scripts with their own configs and look like abandoned / superseded eval-suite branches (ACES composite → EQ task translation, then per-code aggregation).

Verification:
\`\`\`console
$ grep -rn "from every_query.\(aces_to_eq\|process_composite\)\|import .*aces_to_eq\|import .*process_composite" .
# no results
\`\`\`

## Question for @mmcdermott

You're the last author on both (d607a1f). Before I rip them out for #59:

1. Are either of these modules part of a planned but not-yet-wired pipeline (e.g., ACES-composite-task → per-code eval comparison workflow), or are they leftovers from a superseded eval approach now replaced by `eval_suite/`?
2. If they're live, what are the invoking shell commands / notebooks that should be exercising them? (If it's external, I'd like to either promote them to `EQ_*` entry points or document the invocation.)
3. If they're dead: OK to remove in the doc-overhaul PR?

## Proposed resolution (pending your answer)

- **If live, undocumented**: promote to `EQ_aces_to_eq` / `EQ_process_composite` entry points and add to README pipeline section.
- **If dead**: remove the modules and their configs; remove `aces_to_eq` / `process_composite` glob entries from packaging.

cc @payalchandak

### LINKED ISSUE #66: `eval_config.yaml` hardcodes stale dated `model_run_dirs` — eval can't run without editing the shipped config

### LINKED ISSUE #67: `upload_models.py` is package-internal but only invoked via a Columbia-cluster SLURM wrapper

### LINKED ISSUE #79: Phase 1: Restructure core pipeline into prepare_tasks/ and pretrain/ submodules
Parent: #54 (Phase 1)

## Goal

Reorganize the two central pipeline stages into per-stage submodules so each stage owns its code + configs + entry point in one place. Mirrors MEDS-EIC-AR's layout and makes it obvious which files belong to which pipeline step.

## Moves

| Before | After |
|---|---|
| `src/every_query/sample_tasks.py` | `src/every_query/prepare_tasks/__main__.py` |
| `src/every_query/sample_tasks_config.yaml` | `src/every_query/prepare_tasks/configs/sample_tasks_config.yaml` |
| `src/every_query/train.py` | `src/every_query/pretrain/__main__.py` |
| `src/every_query/config.yaml` | `src/every_query/pretrain/configs/config.yaml` |
| `src/every_query/_demo_train.yaml` | `src/every_query/pretrain/configs/_demo_train.yaml` |

## Entry-point rename

```toml
[project.scripts]
EQ_prepare_tasks = "every_query.prepare_tasks.__main__:main"  # was EQ_generate_tasks
EQ_pretrain      = "every_query.pretrain.__main__:main"        # was EQ_train
```

Open decision: transitional aliases for one release (`EQ_generate_tasks` / `EQ_train` keep working, emit `DeprecationWarning`), or hard break? Default proposal: **hard break**, since the 7 existing CLIs are still unstable and no external user is pinning a name yet.

## Imports + references to update

- `importlib.resources.files("every_query")` paths referenced in `__main__.py` files.
- Hydra `config_path` / `config_name` arguments.
- Any test that does `from every_query.sample_tasks import …` → `from every_query.prepare_tasks import …`, same for `train`.
- `conftest.py` docstring + fixture wiring.
- Smoke test parametrization in `tests/test_cli_smoke.py`.

## Scope exclusions

- No behavior changes. This PR is a mechanical move. Tests should continue to pass bit-identically.
- No schema work (that's #80).
- No `predict/` or `evaluate/` stand-up (those are #81 / #83).
- No preprocessing changes (PR #74).

## Acceptance criteria

- [ ] `pytest tests/ src/` passes unchanged count.
- [ ] `EQ_prepare_tasks --help`, `EQ_pretrain --help` work after `pip install`.
- [ ] Old entry-point names are removed (or aliased with a deprecation, depending on the decision).
- [ ] README + eval_suite README updated if they reference the old names.

## Why first

Every Phase 2 PR would otherwise drag both old-and-new layouts; landing this first means each downstream PR touches exactly the submodule it belongs to.

### LINKED ISSUE #80: Phase 2.1 (design): Define cross-stage task-query schema via flexible_schema

### LINKED ISSUE #81: Phase 2.2: Add EQ_predict inference-only entry point

### LINKED ISSUE #91: train.py: do_resume=True should reject structural config drift from the resumed-from run

### LINKED ISSUE #130: aces_to_eq: resolve unused `target_rows` parameter
`create_eq_task_df` in `src/every_query/predict/external_tasks/aces_to_eq.py` accepts a `target_rows` parameter, and the Hydra config plumbs `target_rows_per_shard` through to it, but the parameter is never used inside the function. The aces-to-eq conversion currently emits the full joined shard per code with no row-limiting step.

This was flagged by Copilot during review of #96 (see https://github.com/payalchandak/EveryQuery/pull/96#discussion_r3107684359) and the original scaffolding of this field in 4d7f974 did not wire it up either. It is pre-existing and out of scope for the schema refactor PR.

## Options

1. **Remove**. Drop `target_rows` from the function signature + `aces_to_eq.yaml` config + any entry-point wiring. Matches what the code actually does today (no sampling).
2. **Implement**. Add a sampling/truncation step before the per-code writes so the emitted shards honor `target_rows_per_shard`. Useful if we want balanced shard sizes for evaluation.

Default recommendation is option 1 unless there is a concrete need for row-limiting in the aces eval pipeline — "remove the dead knob" is the simpler + less surprising outcome.

Refs #96.

mmcdermott:
# Release candidate — `dev` → `main`

Rolls up the refactor work tracked under #54 since the last main cut.  ~30 PRs, organised below by the phases from #54's plan.  CI green on every constituent PR; `dev` is green on HEAD.

## Known release-surface caveat (not a blocker; worth calling out up-front)

`EQ_evaluate` is in a **dual-path state** in this release:

- The new consolidated evaluator (`every_query.evaluate.evaluate.main`, landed in #100) is the Phase 2.4 target shape — consumes a `PredictionSchema` parquet from `EQ_predict` and writes per-`(query, duration_days)` metrics.  It's reachable today as `python -m every_query.evaluate.evaluate predictions_parquet=... metrics_parquet=...`, and it's what the slow training-validity E2E (`tests/training_validity/`) already uses end-to-end.
- The `[project.scripts]` `EQ_evaluate` console script still points at the legacy four-stage `every_query.evaluate.eval:main` until the consolidation-wave rewire lands.  See `evaluate/README.md` for the full dual-path explanation.

Users installing this release get the new module but have to invoke it via `-m` until the rewire (tracked on #83, along with the companion legacy-file deletions and `gen_index_times` / `select_model` relocations).  Nothing's broken — every test and both paths point at code that works — but the CLI story isn't fully aligned yet, so this release is better thought of as "Phase 1 + 2.1 + 2.2 + partial 2.4" rather than a clean end-to-end refactor.

### Console-script entry points (#61)
Registered `EQ_process_data`, `EQ_generate_tasks`, `EQ_train`, `EQ_evaluate`, `EQ_predict`, `EQ_gen_eval_index`, `EQ_gen_eval_tasks`, `EQ_select_model` via `[project.scripts]`.  `importlib.resources.files()` used for config paths so a fresh `pip install` resolves them.  `EQ_aces_to_eq` + `EQ_process_composite` are **not** in this release candidate — they're on draft PR #95 (closes #62) and will flow through a later release.

### Phase 1 — submodule reorg (#90, closes #79, subsumes #59/#69)
Every module moved into a role-based submodule: `preprocessing/`, `generate_tasks/`, `train/`, `predict/`, `evaluate/`, `data/`, `model/`, `utils/`, `paper_experiments/`.  Top-level README rewritten from scratch to reflect the final layout.  `paper_experiments/` scaffolded (#89).

### Phase 1 — preprocessing endpoint (#74)
`EQ_process_data` lands.  Pre-MTD transforms + MTD tensorization behind a single entry point.

### Phase 2.1 — `TaskQuerySchema` (#96, closes #80 + #122)
Cross-stage schema contract.  `TaskQuerySchema(LabelSchema)` with `query: pa.large_string` + `duration_days: pa.float32` + nullable `boolean_value` override (null=censored, True=occurred, False=no-event-not-censored).  Sampler, dataset, external-tasks pipelines all driven off exported `{col}_name` / `{col}_dtype` attributes + `align()` at write boundaries.  Collapses the old `(boolean_value, occurs)` pair into a single nullable column.

### Phase 2.2 — `EQ_predict` (#99, closes #81)
Inference-only entry point.  `tasks_dir` → `PredictionSchema` parquet.  Single order-preserving pass through the dataset; `setup_model` extracted to `every_query.utils.model_loader`; `PredictionSchema(TaskQuerySchema)` adds `censor_prob` + `occurs_prob`.  Subprocess-tested E2E; the slow training-validity test (#119) now runs predictions through `EQ_predict`.

### Phase 2.4 — `EQ_evaluate` new shape (#100, partial #83)
`every_query.evaluate.evaluate.main` consumes `PredictionSchema` parquets, emits per-`(query, duration_days)` metrics (`n_rows`, `n_occurs_labeled`, `n_positive`, `occurs_auroc`, `censor_auroc`).  ~40 lines vs. 330 in the legacy `eval.py`.  Reachable via `python -m` — the `[project.scripts]` rewire + legacy-file deletions are deferred to the consolidation wave (still tracked on #83).

## Training stack

- **Train CLI** — resume-path fixes (#94 closes #91, #86 closes #31), RNG-seed-before-instantiate (#124).
- **Sampler** — pure function + removal of caching (#116); schema-driven output (#96).
- **Model / loss** — NaN-loss fix when every sample in a batch is censored (#34).

## Test suite — E2E integration path (#54 Phase 5)

- **Foundation** — CLI-chain fixtures + smoke (#110, #111, #112, #113).
- **Training-validity E2E** — Design 2 oracle markers, asserts both heads learn + unified AUROC criteria (#119, refined through #124).  Now runs inference via `EQ_predict` (#99).
- **CLI smoke** (#110) — every `EQ_*` console script resolves via `--help`.
- **Resume differential** — config mutation detection (#94, #113).

## Developer tooling

- Python minimum bumped to 3.11 (#50).
- `meds-torch-data` to `~=0.7.0` (#73).
- pre-commit plugin pinning (#70, #71).
- codecov-action v5 + relaxed `fail_ci_if_error` (#121).
- Removed dead code: `tasks.py` + `collate_tasks` (#76 closes #53), `upload_models` (#77 closes #67), dup `_env.py` + unused `lit_datamodule.py` (#93), `derive_seed` moved to `utils.seeds` (#78 closes #52).

## Config / env cleanup

- `eval_config.yaml` — dropped stale dated `model_run_dirs` paths (#126 closes #66's first half).
- `PROCESSED` / `INTERMEDIATE` env-var theatre dropped (#127).
- `_collect_codes` resilient to `None` buckets (#87 closes #32).

## Docs

- Top-level README rewritten (#90).
- Subsequent reconciliation pass after Phase 2.1/2.2 landed (#128).

## Post-merge cleanup (#132 + #133, also in this release)

- Doctests: replaced `assert`-style shape checks with show-value (`TaskQuerySchema`, `empty_task_query_df`, `predict_step`).  Added doctests to `_validate_tasks_dir` + `_warn_out_of_vocab` covering happy + error branches.
- Schema-driven names: replaced hardcoded `"boolean_value"` / `"query"` / `"duration_days"` in `data/dataset.py` with `TaskQuerySchema.{col}_name`.
- `utils/model_loader`: `trainer.logger = ""` → `False` (Lightning-typed).
- `predict/external_tasks/aces_to_eq`: unused `target_rows` parameter + `target_rows_per_shard` config knob dropped (closes #130); `assert` sanity check upgraded to a `ValueError` so `python -O` can't strip it.
- `predict/README.md` layout is now a live `print_directory` doctest; can't silently drift from the filesystem.
- `evaluate/README.md` rewritten to document the dual-path state (new `evaluate.py` via `python -m` vs. legacy `eval.py` under `EQ_evaluate` pending the #83 rewire).
- `test_training_validity.py` — wires `every_query.evaluate.evaluate` into the slow E2E; now both `EQ_predict` and `evaluate.evaluate` run as subprocesses against the actual CLI surface, AUROC thresholds land on the metrics-parquet columns.
- `train/train.py`: stray `print("fitting model")` → `logger.info("Fitting model")` (#133), Copilot-flagged during the #131 re-review.

## Still open for a follow-up release

- **#82 / #83 consolidation wave**: `EQ_evaluate` entry-point rewire + delete `evaluate/eval.py`, `evaluate/gen_task.py`; move `select_model.py` → `paper_experiments/leaderboard/`; fold `gen_index_times.py` into `generate_tasks/`; drop `model_run_dirs` list config (closes rest of #66).
- **#85** — dataset-agnostic `sample_codes` rewrite (PR #97 open).
- **#62** — external-task entry points (PR #95 open).
- **#129** — `occurs_prob` → `label_prob` rename post-NeurIPS.

## Test plan for this merge

- [x] CI green on `dev` HEAD (all constituent PRs passed individually, including the #132 cleanup).
- [x] `uv run pytest -q` — 214 passed, 1 deselected (slow) locally.
- [x] Slow training-validity E2E passes locally via `EQ_predict` → `evaluate.evaluate` subprocess chain (~8 min).
- [ ] Full `pytest -m 'slow or not slow'` on a clean checkout of this merge commit.
- [ ] Manual spot-check of `EQ_*` console scripts on a real cohort.


src/every_query/utils/model_loader.py
```diff
@@ -0,0 +1,87 @@
...
+        NotADirectoryError: if ``model_run_dir`` isn't a directory.
+        FileNotFoundError: if no checkpoint can be resolved.
+    """
+    model_run_dir = Path(model_run_dir)
+    if not model_run_dir.is_dir():
+        raise NotADirectoryError(f"{model_run_dir} is not a directory")
+
+    train_cfg = OmegaConf.load(model_run_dir / "resolved_config.yaml")
+    train_cfg.trainer.logger = ""
+
```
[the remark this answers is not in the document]
mmcdermott: Good catch — fixed in #132 (commit 4ce840a).  `train_cfg.trainer.logger = False` now (was `""`, which worked by coincidence since it's falsy, but isn't in Lightning's `Trainer(logger=...)` type contract).  Added a comment explaining the purpose: avoid re-initing the training-time wandb run during predict/eval.

Note: this line is on `every_query.utils.model_loader.py`, landed in #99 — so the fix rides through #132 → dev → this release-candidate branch on the next refresh.

pyproject.toml
```diff
@@ -55,7 +55,22 @@ every_query = ["**/*.yaml"]
...
-[project.scripts] # Add any console scripts here
+[project.scripts]
+EQ_process_data = "every_query.preprocessing.__main__:main"
+EQ_train = "every_query.train.train:main"
+EQ_evaluate = "every_query.evaluate.eval:main"
+EQ_generate_tasks = "every_query.generate_tasks.sample_tasks:main"
+EQ_gen_eval_index = "every_query.evaluate.gen_index_times:main"
+EQ_gen_eval_tasks = "every_query.evaluate.gen_task:main"
+EQ_select_model = "every_query.evaluate.select_model:main"
+EQ_predict = "every_query.predict.predict:main"
```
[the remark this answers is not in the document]
mmcdermott: Fair — the PR body oversold what's actually in the release-candidate tree.  `EQ_aces_to_eq` / `EQ_process_composite` are on draft PR #95 (closes #62) and not yet merged to dev.

Fixed the PR description to list only what's actually in `[project.scripts]` here (EQ_process_data, EQ_generate_tasks, EQ_train, EQ_evaluate, EQ_predict, EQ_gen_eval_index, EQ_gen_eval_tasks, EQ_select_model) and explicitly call out that the two aces/composite entry points are deferred to a later release.

src/every_query/predict/external_tasks/aces_to_eq.py
```diff
@@ -0,0 +1,127 @@
...
+
+
+def create_eq_task_df(
+    eq_shard_fp: str,
+    aces_shard_fp: str,
+    codes: list[str],
+    output_root: str,
+    target_rows: int,
+    duration_days: float,
+) -> int:
```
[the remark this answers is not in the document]
mmcdermott: Already tracked — filed as #130 during the #96 review when this came up the first time (see the discussion on that issue for the two options: remove the parameter or wire it to a real sampling step).  Out of scope for this release-candidate PR; will land in whichever direction #130 resolves to.

src/every_query/predict/external_tasks/aces_to_eq.py
```diff
@@ -0,0 +1,127 @@
...
+    aces_shard_df = aces_shard_df.rename({"boolean_value": "task_label"})
+    eq_shard_all_tasks_df = pl.read_parquet(
+        eq_shard_fp, columns=["subject_id", "prediction_time", "censored", *codes]
+    )
+    shard_name = Path(eq_shard_fp).name
+
+    joined_df = aces_shard_df.join(eq_shard_all_tasks_df, on=["subject_id", "prediction_time"], how="left")
+
+    assert joined_df.select(pl.col("censored").null_count()).item() == 0
+
```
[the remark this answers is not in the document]
mmcdermott: Valid point — `assert` gets stripped under `python -O`, which would silently let a malformed join through.  Fixed in #132 (commit 4ce840a) — the check now raises a `ValueError` with the null row count + both shard paths so the failure is actionable regardless of Python's optimization flag.

The fix rides into dev via #132, then into this release candidate on the next refresh.

mmcdermott:
@copilot review

mmcdermott:
Thanks for the re-review — summary noted.

**On the deferred items**: minor correction — `target_rows` / #130 is actually closed already (removed in #132, merged into dev before the re-trigger).  The other two (#82/#83 consolidation wave, #129 label rename) are indeed still open and tracked.

**On the `print("fitting model")` finding**: good catch — the fix was one line so I went with the "one-line fix in a follow-up" path rather than filing a ticket.  See #133, which targets dev and will flow through on the next release candidate refresh.  Couldn't justify scope-expanding #131 itself to include it since that'd mean rebasing the merge PR.

mmcdermott:
some bigger picture comments from chatgpt: 
My take: I would **not block this PR on code correctness** based on the actual paths I checked, but I **would call out one important release-surface caveat**.

The PR is very large for a merge roll-up — **192 commits, 102 changed files, +5123/-2608** — and it is intended to land the schema-driven preprocess → generate_tasks → train → predict → evaluate flow plus broader subprocess/E2E coverage. ([[GitHub](https://github.com/payalchandak/EveryQuery/pull/131)][1]) 

What looks good:

* The new **nullable `boolean_value` contract** appears internally consistent across sampler output, dataset collapse into `(censor, occurs)`, reconstruction in `predict`, and filtering in `evaluate.metrics`.
* The new tests are materially better than smoke-only coverage: `EQ_predict` has an order-preservation test, `EQ_generate_tasks` recomputes labels from raw event shards, and the slow training-validity test now actually runs the predict → evaluate subprocess chain.
* The earlier plausible blockers that *would* have worried me were already addressed in follow-up discussion: the `trainer.logger=""` typing issue was fixed to `False`, and the production `assert` in `aces_to_eq.py` was replaced with a real exception path. 

My main caveat:

* The shipped CLI surface is still a little incoherent. The PR adds the new consolidated evaluator, but **`EQ_evaluate` still points at the legacy `evaluate/eval.py` path**, while the new evaluator is exercised via `python -m every_query.evaluate.evaluate`; that deferral is explicitly tracked in the follow-up consolidation wave.  

So the comment I’d leave is basically:

> I don’t see a blocker in the new schema-driven path itself. The nullable-label contract looks consistent end-to-end, and the new subprocess/E2E tests are doing real work here rather than just smoke testing.
>
> My main concern is release-surface coherence rather than correctness: this RC ships the new consolidated evaluator, but `EQ_evaluate` still points to the legacy path, so users don’t yet get a fully aligned CLI story. Since that’s already tracked in #82/#83, I’m okay with this as a release-candidate roll-up — just worth being explicit that evaluation is still in a dual-path state.

Net: **looks mergeable to me**, with that caveat.

[1]: https://github.com/payalchandak/EveryQuery/pull/131 "Release candidate — refactor Phase 1 + 2.1/2.2 + partial 2.4 (roll-up) by mmcdermott · Pull Request #131 · payalchandak/EveryQuery · GitHub"

mmcdermott:
Agreed with the ChatGPT take — no code-correctness change warranted, and the dual-path `EQ_evaluate` caveat was being under-sold.  The PR description had it in the "Still open" section at the bottom; I've hoisted it to a prominent "Known release-surface caveat" section right after the opener so it's impossible to miss when scanning the PR.

Also refreshed the PR body to:
- Include #133 (the stray `print → logger.info` fix just merged into dev) alongside #132 in the post-merge-cleanup section.
- Drop the stale #130 reference from the follow-up list (closed in #132).

Title stays as-is — "Phase 1 + 2.1/2.2 + partial 2.4 (roll-up)" already advertises the partial 2.4 scope, which is the coherence gap ChatGPT flagged.

mmcdermott:
@payalchandak @gkondas — before merging this release, could I get your input on a few things that need workflow context I don't have?

## 1. The `EQ_evaluate` dual-path state — is it worth fixing *here* rather than waiting on #83?

Current release ships:
- **New**: `every_query.evaluate.evaluate.main` — consumes a `PredictionSchema` parquet from `EQ_predict`, emits per-`(query, duration_days)` metrics.  Reachable via `python -m every_query.evaluate.evaluate`.  This is what the training-validity E2E uses.
- **Legacy**: `every_query.evaluate.eval.main` — loads a model, runs inference itself (no `EQ_predict` dependency), does multi-model / id-ood-manual orchestration, writes per-code AUCs.  `[project.scripts]` `EQ_evaluate` still points here.

These aren't equivalent — the legacy does orchestration the new module doesn't.  Options:

- **(a) Ship as-is** (current plan): `EQ_evaluate` console script stays legacy; users reach the new module via `python -m`.  Dual-path tracked on #83.  No breaking change.
- **(b) Rewire now**: flip `EQ_evaluate` → new `evaluate.evaluate:main`.  Breaks anything that invokes the console script expecting the multi-model / id-ood-manual behavior.  Want to flag this up-front rather than surprise a user.
- **(c) Rename the legacy**: keep both but rename the legacy entry to `EQ_evaluate_legacy` and point `EQ_evaluate` at the new module.  Soft deprecation — preserves the legacy flow for callers who want it, surfaces the new one as the default.

**My lean**: (a), since the E2E path actually *works* (the training-validity test proves it), and (b)/(c) would mean choosing names / deprecation timing without knowing what the paper experiments pipeline is actually calling today.  But if you'd rather take the hit now to ship a clean CLI story, I can do (c) as a small follow-up PR into dev before this RC merges.

**Question**: which option do you prefer?  Concretely: does your actual paper-experiments workflow still call `EQ_evaluate` (legacy behavior), or has it already switched to `EQ_predict` + `python -m every_query.evaluate.evaluate`?

## 2. Manual spot-check on a real cohort — any volunteers?

The RC test plan has `[ ] Manual spot-check of EQ_* console scripts on a real cohort` unchecked.  I've only run things against the synthetic Design-2 dataset.  Before cutting the release tag, worth having one of you run `EQ_process_data` → `EQ_generate_tasks` → `EQ_train` → `EQ_predict` against an actual cohort to catch anything the CI tests don't.  Happy to queue it myself if you point me at a cohort path, but you two are closer to what "normal use" looks like.

## 3. Other follow-ups that could land before vs. after the RC cut

- #95 (EQ_aces_to_eq + EQ_process_composite entry points) — still draft; should it flow into this RC or the next?
- #97 (dataset-agnostic `sample_codes`) — same question.
- #129 (`occurs_prob` → `label_prob` rename) — confirmed post-NeurIPS, or is there a window to slip it in?

All of these are "is it worth blocking the release vs. shipping as-is," which depends on what downstream is waiting for a tagged release.

No rush — just want the right calls before we merge.

gkondas (commented):
I want to switch to using the new staging which separates predict and evaluate into two separate end points. However, we still need a way to generate evaluation task dataframes and labels for EQ_predict to consume. I think we want a new endpoint of something along the lines of EQ_gen_eval_tasks which is distinct from the current generate_tasks endpoint as described below.

During pretraining we want pretraining df's where the rows can each be a unique task, i.e. (code x duration_days) can be different in each row. This is how generate_tasks is currently setup and makes sense for generating pretraining tasks/labels. However, for evaluation of a task on the entire held_out split, we need to generate a task df for each task where each (subject_id,prediction_time) in the split has a row for each (code x duration) that the user wants to evaluate the model on. In the old pipeline, these evaluation task dataframes were generated by running gen_index.py and gen_task.py in sequence, then poining eval.py at these outputs. However, in the new staging I think it makes sense for an explicit endpoint which generates evaluation task dataframes where the user can specify how many tasks they want to evaluate on, and the end point samples that many tasks (similar to the new everyquery.generate_tasks endpoint) and writes each of these task df's to disk. Then every_query.predict can be pointed at these evaluation task dataframes to produce predictions on these tasks that can later be consumed by every_query.evaluate. In my opinions, generating evaluation dataframes is different enough than generating pretaining dataframes and should have its own endpoint. Maybe generate_tasks can be renamed to generate_training_tasks to make clear it is different than a new generate_evaluation_tasks endpoint.

mmcdermott:
@gkondas thanks — this is a real gap and worth planning carefully.  @payalchandak tagging you since this affects the refactor plan.

## What Greg is pointing at

The new pipeline (`EQ_generate_tasks` → `EQ_train` → `EQ_predict` → `every_query.evaluate.evaluate`) is complete for **pretraining-style workflows** where each row is an independent `(subject, prediction_time, query, duration_days)` sample.

It's **not** complete for **evaluation-style workflows**, which need a dense grid: for every `(subject, prediction_time)` in the held-out split, one row per `(query × duration)` combination the user wants AUCs for.  The legacy pipeline produced this via `EQ_gen_eval_index → EQ_gen_eval_tasks`; the new pipeline has no equivalent.

The training-validity slow test hides this gap because it hand-builds the evaluation task parquet in `_compute_labels`.  A user starting from a real cohort + trained checkpoint today could not produce evaluation predictions through `EQ_predict` without either (a) writing the task parquet themselves, or (b) falling back to the legacy `EQ_evaluate` path (which runs its own inference and bypasses `EQ_predict` entirely).

So the new pipeline *runs* end-to-end for evaluation in tests but isn't usable end-to-end for a real caller.  That's a bigger coherence issue than just the `EQ_evaluate` console-script rewire.

## Revision options for this RC

### A.  Doc-only: acknowledge the gap, don't block the release  ← **my recommendation**

- Update #131's "Known release-surface caveat" to call out the evaluation-task-generation gap explicitly, alongside the `EQ_evaluate` rewire deferral.  A user reading the PR body would know the new pipeline is "pretraining-complete" but not "evaluation-complete" yet.
- File a new issue (tentatively `Phase 2.5 — evaluation-task-generation endpoint`) pinning the scope: produce dense-grid `TaskQuerySchema` parquets for held-out evaluation, matching Greg's proposed semantics.  Becomes the tracking ticket for the follow-up PR.
- Ship this RC.  Paper-experiments workflow stays on legacy `EQ_evaluate` for held-out evaluation until the new endpoint lands; any pretraining workflow uses the new stack end-to-end.

### B.  Rename + doc: set up for the follow-up without writing the new endpoint

Everything in (A), plus:
- Rename `EQ_generate_tasks` → `EQ_generate_training_tasks` (the current endpoint only produces pretraining-shape tasks).  Console script is added, old name kept as a deprecated alias for one release so existing callers don't break.
- Rename `src/every_query/generate_tasks/` → `src/every_query/generate_training_tasks/` (or keep the dir, just retarget the CLI name).

Creates conceptual room for `EQ_generate_evaluation_tasks` to land next without a confusing "which `generate_tasks` does what" moment.  But it's churn for the sake of future clarity — if (A)'s doc update is clear enough, this is deferrable.

### C.  Land the new endpoint in this RC

Everything in (B), plus:
- Add `EQ_generate_evaluation_tasks` with Greg's proposed semantics: user specifies the split, number of prediction times per subject, and `{code_group, duration_grid}`.  Endpoint samples prediction times (via the logic inherited from `gen_index_times.py`), materializes the dense grid, writes a single `TaskQuerySchema` parquet.
- Tests: unit (sampling determinism, grid shape) + E2E (`EQ_generate_evaluation_tasks` → `EQ_predict` → `evaluate.evaluate` on the synthetic cohort).
- New path becomes usable end-to-end for evaluation, closing the coherence gap completely.

Substantial scope — probably 2-3 days between design, impl, test.  Blocks the RC for that long.  Decision depends on how urgently downstream needs a tagged release.

## Questions for the team before I start any of these

1. **Which option — A, B, or C?**  My lean is A + file the tracking issue; consider B as a low-churn follow-up that unblocks C.  If downstream is waiting for a tagged release, strongly lean A.  If the release can slip, C is the most complete story.

2. **Greg's renaming suggestion — `generate_tasks` → `generate_training_tasks`?**  I think yes, but deliberately as a separate follow-up from both the new endpoint and the RC itself.  Easier to review a pure rename than a rename bundled with a new feature.

3. **Does `EQ_generate_evaluation_tasks`'s task-sampling logic reuse / supersede the legacy `gen_index_times.py`?**  I'd prefer to pull the deterministic prediction-time sampler out of `gen_index_times.py` into a shared util and have both the new and legacy endpoints call it.  That keeps the legacy path functional until we retire it and doesn't duplicate sampling logic.  But that's a scoping decision that intersects #82/#83's consolidation wave.

4. **Payal — does the paper-experiments code assume the legacy `EQ_gen_eval_index + gen_task + EQ_evaluate` path today?**  If yes, option A's "new pipeline is pretraining-complete, legacy path still supported for evaluation" is a reasonable interim; option B's rename wouldn't affect your scripts (the evaluation side stays on legacy).  If no (you've already switched to `EQ_predict` + the new module), then the gap Greg is pointing at is probably blocking you too and option C is worth waiting for.

Will plan further once you both weigh in.  In the meantime I'm not making any changes to this PR.  Ready to execute whichever option we land on.

src/every_query/data/schema.py
```diff
@@ -0,0 +1,136 @@
...
+        0
+        >>> for name, dtype in df.schema.items():
+        ...     print(f"{name}: {dtype}")
+        subject_id: Int64
+        prediction_time: Datetime(time_unit='us', time_zone=None)
+        query: String
+        duration_days: Float32
+        boolean_value: Boolean
+    """
+    return pl.DataFrame(
```
payalchandak: This function manually mirrors the PyArrow types from TaskQuerySchema above. What happens if these two type definitions can silently drift? Consider deriving the Polars schema programmatically from the TaskQuerySchema fields so there's a single source of truth.

src/every_query/data/dataset.py
```diff
@@ -29,8 +31,6 @@ class EveryQueryBatch(MEDSTorchBatch):
     """
 
     # Extra task annotations (subject-level)
-    subject_id: torch.LongTensor | None = None
-    prediction_time: torch.LongTensor | None = None
```
payalchandak, line 33: Why was this deleted? This existed so we can know which subject/time index a sample belongs to at inference time

src/every_query/data/dataset.py
```diff
@@ -324,21 +324,54 @@ def get_task_seq_bounds_and_labels(cls, label_df: pl.DataFrame, schema_df: pl.Da
...
-        self.has_occurs: bool = "occurs" in self.schema_df.collect_schema().names()
-        self.has_query: bool = "query" in self.schema_df.collect_schema().names()
-        self.has_duration_days: bool = "duration_days" in self.schema_df.collect_schema().names()
+        self.has_occurs: bool = "occurs" in schema_cols
+        self.has_query: bool = TaskQuerySchema.query_name in schema_cols
+        self.has_duration_days: bool = TaskQuerySchema.duration_days_name in schema_cols
         self.occurs = self.schema_df["occurs"] if self.has_occurs else None
-        self.query = self.schema_df["query"] if self.has_query else None
-        self.duration_days = self.schema_df["duration_days"] if self.has_duration_days else None
+        self.query = self.schema_df[TaskQuerySchema.query_name] if self.has_query else None
```
payalchandak: should this not be `has_code` as we are thinking of the query as comprising of a code and duration and this field is only about the code

src/every_query/predict/schema.py
```diff
@@ -0,0 +1,68 @@
...
+
+from every_query.data.schema import TaskQuerySchema
+
+
+class PredictionSchema(TaskQuerySchema):
+    """EveryQuery prediction-output row.
+
+    Inherits every column from :class:`every_query.data.schema.TaskQuerySchema` —
+    ``subject_id``, ``prediction_time``, ``query``, ``duration_days``, plus the inherited
+    optional label columns (``boolean_value`` nullable, per ``LabelSchema``).  Adds:
```
payalchandak: Should be nullable, per ``TaskQuerySchema``

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,315 @@
...
+    if not parquets:
+        raise ValueError(f"tasks_dir {tasks_dir} contains no parquet files")
+    if len(parquets) > 1:
+        logger.warning(
+            f"tasks_dir {tasks_dir} contains {len(parquets)} parquet files; all will be "
+            f"concatenated.  The typical inference use case is a single flat parquet."
+        )
+
+
+def _warn_out_of_vocab(task_codes: set[str], train_cfg: DictConfig) -> None:
```
payalchandak: This should throw errors and not run prediction rather than warn and output garbage probabilities

src/every_query/predict/predict.py
```diff
@@ -0,0 +1,315 @@
...
+    missing = task_codes - training_vocab
+    if missing:
+        logger.warning(
+            f"{len(missing)} of {len(task_codes)} task-query codes are not in the model's training "
+            f"vocabulary and will be PAD-encoded (predictions will be near-uniform for these rows): "
+            f"{sorted(missing)[:10]}{'...' if len(missing) > 10 else ''}"
+        )
+
+
+def _gather_probabilities(pred_batches: list[dict[str, torch.Tensor]]) -> pl.DataFrame:
```
payalchandak, line 197: This function should read the column type (e.g. pl.Float32) from the definition in `src/every_query/predict/schema.py`

src/every_query/generate_tasks/README.md
payalchandak, line 1: **Naming is inconsistent across three layers**, and the two sibling commands don't agree either:

| Layer | Training | Evaluation |
|---|---|---|
| CLI | `EQ_generate_training_tasks` | `EQ_generate_evaluation_tasks` |
| Module | `sample_tasks.py` | `sample_evaluation_tasks.py` |
| Config | `sample_tasks_config.yaml` | `sample_evaluation_tasks_config.yaml` |

Two issues:
1. CLI verb (`generate`) ≠ file verb (`sample`), so grepping for `generate_training` from the CLI name finds nothing — `pyproject.toml` entry points are the only bridge.
2. Training drops the qualifier, evaluation keeps it (`sample_tasks.py` vs. `sample_evaluation_tasks.py`), making the training module read as the generic default.

Suggested rename so all three layers match the CLI:
- `sample_tasks.py` → `generate_training_tasks.py`
- `sample_evaluation_tasks.py` → `generate_evaluation_tasks.py`
- (+ matching config filenames)

src/every_query/generate_tasks/sample_evaluation_tasks.py
```diff
@@ -0,0 +1,389 @@
...
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Pure primitives
+# ---------------------------------------------------------------------------
+
+
+def sample_prediction_times_per_subject(
```
payalchandak, line 54: not a codebase concern, but @gkondas are we currently sampling K times per subject rather than taking all? is this also what ETHOS does?

src/every_query/generate_tasks/sample_evaluation_tasks.py
payalchandak, line 1: The eval task grid is codes × durations, not (code, duration) tuples.

You can't specify individual (query, duration) pairs today. The config takes two separate lists (codes and durations), and `build_evaluation_index_df` in `sample_evaluation_tasks.py` cross-joins them into the full grid.

So `codes: [HR, TEMP]` with `durations: [7, 30]` always produces all four combinations: (HR, 7), (HR, 30), (TEMP, 7), (TEMP, 30) — there's no way to restrict to just (HR, 7) and (TEMP, 30).

Both modes should eventually exist: the full cross-join (useful for per-task eval metrics over a grid) and an explicit list of (code, duration) tuples (useful when the code universe is large and you only care about specific pairs). Could happen now or file an issue as future work.

src/every_query/predict/README.md
payalchandak, line 1: EQ_predict will overwrite an existing predictions parquet every time. There's no existence check on output_parquet before running?

src/every_query/utils/_env.py
```diff
@@ -13,11 +13,15 @@
...
-    "INTERMEDIATE",
     "FINAL_DATA_DIR",
     "WANDB_ENTITY",
 )
+# `PROCESSED` and `INTERMEDIATE` used to live here too but no Hydra config
+# interpolates `${oc.env:PROCESSED}` or `${oc.env:INTERMEDIATE}`.  Their only use is a
+# dotenv fallback inside `sample_tasks.py::_resolve_path`, which already tolerates
+# missing env vars when config values are supplied (the normal CLI path).  Keeping
+# them in the gate was pure friction: demo fixtures / `--help` invocations had to
+# set throwaway placeholder values for no actual resolution.  See #117.
```
payalchandak, line 29: PROCESSED and INTERMEDIATE were used for preprocessing

src/every_query/model/lightning_module.py
```diff
@@ -286,37 +287,48 @@ def test_step(self, batch: EveryQueryBatch) -> torch.Tensor:
     def predict_step(self, batch: EveryQueryBatch) -> dict[str, torch.Tensor]:
         """Produce prediction outputs (probabilities, embeddings, labels) for one batch.
 
+        ``query`` and ``duration_days`` ride along from the batch so downstream
+        consumers that inspect predictions at the batch level don't have to re-derive
+        them.  Per-row identifiers (``subject_id`` / ``prediction_time``) are not
+        needed here — ``EQ_predict`` labels output rows by indexing the dataset's
+        ``schema_df`` in dataloader order, so there's no benefit to round-tripping
+        them through a tensor.
```
payalchandak, line 295: Dropping and not joining on the columns (``subject_id`` / ``prediction_time``) requires a guarantee that `shuffle=False` in dataloader, is this enforced?

src/every_query/preprocessing/__init__.py
payalchandak, line 1: @mmcdermott is preprocessing largely duplicating the MEDS EIC AR preprocessing? if so, why have duplicate code here?

payalchandak (approved):
@mmcdermott overally LGTM, I read through all the code in src/ while skipping preprocessing and paper_experiments. Also skipped tests and scripts. Some comments and questions for claude, but nothing blocking

src/every_query/data/schema.py
mmcdermott: Fixed in efbf12e (PR #137 into `dev`). `empty_task_query_df` now builds the frame via `pl.from_arrow(pa.schema(...).empty_table())` using `TaskQuerySchema.{col}_dtype` as the single source of truth — any future change to the PyArrow type declarations flows through automatically.

src/every_query/predict/schema.py
mmcdermott: Fixed in efbf12e (PR #137 into `dev`). Docstring now says "per ``TaskQuerySchema``'s override" rather than "per ``LabelSchema``" — `LabelSchema` declares `boolean_value` non-nullable; `TaskQuerySchema` is the layer that overrides it to nullable (for the censored-null contract from #122).

The actual column declaration was already correct via inheritance — `PredictionSchema(TaskQuerySchema)` picks up the nullable override automatically. This was purely a docstring accuracy fix.

src/every_query/predict/predict.py
mmcdermott: Done in efbf12e (PR #137 into `dev`). `_warn_out_of_vocab` → `_check_vocab`, raises `ValueError` on OOV codes and `FileNotFoundError` if the metadata file is missing. Doctest updated to match the new contract.

Side effect worth flagging: this caught a real test bug. `tests/test_predict_cli.py` was using `HR`/`TEMP` as query codes, but after preprocessing the simple_static_MEDS vocab only has the bucketed forms `HR//value_[...]` — so predictions were effectively PAD-encoded and we never noticed because the old path just warned. Switched to `DISCHARGE`/`DOB` which are real codes in the vocab.

src/every_query/data/dataset.py
mmcdermott: Rationale from commit c6018ec (on #99):

> They were only in the batch so ``predict_step`` could emit them in its output dict, so the old draft could label predictions. The new ``EQ_predict`` flow labels rows by indexing ``D.test_dataset.schema_df`` in dataloader order — no need to round-trip them through a tensor. The model / loss / metrics logging never used them.

Per-sample subject/time info is still available at inference time — `EQ_predict` pulls it straight from `dataset.schema_df` (see `_identifiers_from_schema_df` in `predict.py`) and stitches it back onto the probabilities via an order-preserving hstack. The hstack correctness now has a runtime SequentialSampler assertion too (thread on `lightning_module.py:295`). So the capability is preserved — just not via a tensor round-trip through the batch object.

src/every_query/predict/predict.py
mmcdermott: Agreed — the hardcoded `pl.Float32` in `_gather_probabilities` should be read from `PredictionSchema.censor_prob_dtype` / `occurs_prob_dtype`, same source-of-truth pattern as the schema.py fix in #137. Needs a small PA→PL type helper since `flexible_schema` exposes arrow types but polars needs polars types.

Deferring to a follow-up PR rather than churning #131 further — happy to pick it up after this lands. OK to leave this thread open as a tracked TODO, or would you prefer I file an issue?

src/every_query/data/dataset.py
mmcdermott: Rationale from commit c6018ec (on #99):

> They were only in the batch so ``predict_step`` could emit them in its output dict, so the old draft could label predictions.  The new ``EQ_predict`` flow labels rows by indexing ``D.test_dataset.schema_df`` in dataloader order — no need to round-trip them through a tensor.  The model / loss / metrics logging never used them.

Per-sample subject/time info is still available at inference time — `EQ_predict` pulls it straight from `dataset.schema_df` (see `_identifiers_from_schema_df` in `predict.py`) and stitches it back onto the probabilities via an order-preserving hstack.  The hstack correctness now has a runtime SequentialSampler assertion too (thread on `lightning_module.py:295`).  So the capability is preserved — just not via a tensor round-trip through the batch object.

src/every_query/predict/predict.py
mmcdermott: Done in efbf12e (PR #137 into `dev`).  `_warn_out_of_vocab` → `_check_vocab`, raises `ValueError` on OOV codes and `FileNotFoundError` if the metadata file is missing.  Doctest updated to match the new contract.

Side effect worth flagging: this caught a real test bug.  `tests/test_predict_cli.py` was using `HR`/`TEMP` as query codes, but after preprocessing the simple_static_MEDS vocab only has the bucketed forms `HR//value_[...]` — so predictions were effectively PAD-encoded and we never noticed because the old path just warned.  Switched to `DISCHARGE`/`DOB` which are real codes in the vocab.

src/every_query/generate_tasks/README.md
mmcdermott: Agreed on the inconsistency — the two sibling commands not matching is the worst part (asymmetric qualifier drop on training).  Your proposed rename is right:
- `sample_tasks.py` → `generate_training_tasks.py`
- `sample_evaluation_tasks.py` → `generate_evaluation_tasks.py`
- (+ matching config filenames + update `[project.scripts]` module paths)

This is churn across ~10 files (module paths in `pyproject.toml`, imports in dataset.py, 6 test files, both READMEs, both configs).  I'd rather land it as its own focused PR than bundle into #131 at this point — minimizes conflict surface for reviewers and keeps the rename as a reversible atomic unit.

Filing a follow-up issue unless you'd prefer I bundle it here.  Your call.

src/every_query/predict/schema.py
mmcdermott: Fixed in efbf12e (PR #137 into `dev`).  Docstring now says "per ``TaskQuerySchema``'s override" rather than "per ``LabelSchema``" — `LabelSchema` declares `boolean_value` non-nullable; `TaskQuerySchema` is the layer that overrides it to nullable (for the censored-null contract from #122).

The actual column declaration was already correct via inheritance — `PredictionSchema(TaskQuerySchema)` picks up the nullable override automatically.  This was purely a docstring accuracy fix.

src/every_query/predict/predict.py
mmcdermott: Agreed — the hardcoded `pl.Float32` in `_gather_probabilities` should be read from `PredictionSchema.censor_prob_dtype` / `occurs_prob_dtype`, same source-of-truth pattern as the schema.py fix in #137.  Needs a small PA→PL type helper since `flexible_schema` exposes arrow types but polars needs polars types.

Deferring to a follow-up PR rather than churning #131 further — happy to pick it up after this lands.  OK to leave this thread open as a tracked TODO, or would you prefer I file an issue?

src/every_query/generate_tasks/sample_evaluation_tasks.py
mmcdermott: Agreed — the cross-join is the only mode today, and it's the wrong shape when the evaluation set is a hand-picked list of `(code, duration)` pairs rather than a regular grid.

Filed as #138 to track adding an explicit `query_tuples: [{query: HR, duration_days: 7}, ...]` alternative that coexists with the current `codes + durations` grid form.  Deferring to that rather than expanding #131/#135 scope.

src/every_query/model/lightning_module.py
mmcdermott: Good catch — the order-preservation was enforced implicitly by MTD's val/test dataloaders using `SequentialSampler`, but that contract was never asserted on our side.  Added a runtime assertion in efbf12e (PR #137 into `dev`):

```python
sampler_cls = type(getattr(dataloader, "sampler", None)).__name__
if sampler_cls != "SequentialSampler":
    raise RuntimeError(
        f"{dataloader_attr} must use SequentialSampler to preserve row order for the "
        f"predictions hstack; got {sampler_cls!r}.  Re-check the MTD datamodule config."
    )
```

Now a future MTD change (or a datamodule override adding a `RandomSampler`) fails loudly at startup instead of silently corrupting predictions by stitching probs to the wrong identifiers.  The existing `split != "train"` check already blocks the train-split case; this assertion covers the val/test path too.

src/every_query/utils/_env.py
mmcdermott: Good catch — the old comment implied `PROCESSED`/`INTERMEDIATE` were unused, which is wrong.  Fixed the comment in efbf12e (PR #137 into `dev`) to accurately state:

1. Both are used by the preprocessing pipeline (it writes to these dirs) and by `generate_tasks` as `_resolve_path` fallbacks.
2. They're excluded from `REQUIRED_ENV_VARS` only because `_resolve_path` tolerates missing env vars when the CLI supplies the path directly (the normal test + `--help` path).
3. A fresh-machine full preprocessing run should still set both in `.env` — noted explicitly in the comment now.

src/every_query/data/schema.py
mmcdermott: Fixed in efbf12e (PR #137 into `dev`). `empty_task_query_df` now builds the frame via `pl.from_arrow(pa.schema(...).empty_table())` using `TaskQuerySchema.{col}_dtype` as the single source of truth — any future change to the PyArrow type declarations flows through automatically.

src/every_query/preprocessing/__init__.py
mmcdermott: Tagging @mmcdermott for the answer on preprocessing-vs-EIC-AR duplication — this is a scope / strategic call, not something to fix blindly here.

src/every_query/generate_tasks/sample_evaluation_tasks.py
mmcdermott: Noted — this is a design question for @gkondas, no code action from me here.  The `K prediction times per subject` choice comes from the need to cap the dense-grid row count on large cohorts; whether it matches ETHOS's evaluation protocol is exactly the kind of thing that should get pinned down before it diverges silently.

src/every_query/predict/README.md
mmcdermott: Fixed in efbf12e (PR #137 into `dev`).  `EQ_predict` now has an `overwrite: false` config knob (default refuse-to-clobber); re-running against an existing `output_parquet` raises `FileExistsError` at startup unless `overwrite=true` is passed.  Matches the convention used by `EQ_generate_training_tasks` / `EQ_generate_evaluation_tasks`.  README + config YAML updated.

PR #132, opened 2026-04-20
refactor: release-candidate followups — doctests show outputs, schema-driven names, stale docs

mmcdermott:
## Summary

Post-merge cleanup based on a deep review of the #131 release candidate.  Targets five categories; low-risk doctest + docs changes plus a narrow sweep of hardcoded column names that the schema's `{col}_name` attributes already cover.  No behavior change in shipping code.

### Doctests: show values instead of asserting shapes

Where a doctest was `>>> result.shape == (2, 3); True` or `>>> set(cols) == {...}; True`, replaced with the direct `>>> result.shape` / `>>> cols` form so a reader learns the API by seeing the actual output.

- `TaskQuerySchema` — `>>> [f.name for f in aligned.schema]` directly.
- `empty_task_query_df` — `>>> df.columns` + `>>> df.dtypes` directly (previously a set-equality that hid both the column order and the dtypes).
- `EveryQueryLightningModule.predict_step` — show `sorted(result.keys())`, `query_embed.shape`, `(query.shape, duration_days.shape)`; collapse the `>= 0; True` / `<= 1; True` probability-bound check into a single chained comparison that prints the actual bound.

### Schema-driven column names in `dataset.py`

Replaced hardcoded string literals with `TaskQuerySchema.{col}_name` in `EveryQueryPytorchDataset.__init__` + `labels_df`:

- `"boolean_value"` → `TaskQuerySchema.boolean_value_name` on the collapse path and the super-labels re-point.
- `"query"` / `"duration_days"` → `TaskQuerySchema.query_name` / `.duration_days_name` on the `has_*` detection, the `self.query` / `self.duration_days` Series pointers, and the `labels_df` extras list.
- `"occurs"` left as a literal — it's the legacy-shape detection sentinel, not a `TaskQuerySchema` column.

Added the `TaskQuerySchema` import to `dataset.py`.

### Stale docs

- `predict/README.md` — config key `tasks_parquet` → `tasks_dir` (renamed in #99).  Added `split` to the optional-config list.  Updated the pipeline diagram to show a directory of parquets, not a single file.
- `evaluate/README.md` — rewritten to document the dual-path state: the new consolidated `evaluate.py` (from #100) runs today via `python -m`, while the legacy four-stage pipeline is still what the `EQ_evaluate` console script targets until the #83 rewire lands.  Makes the distinction explicit so readers don't get blindsided by the "current" pipeline label drifting off the new entry point.

## Scope kept narrow

Deliberately *not* in this PR (items the review flagged but need more thought):

- Rewiring `[project.scripts]` `EQ_evaluate` → `every_query.evaluate.evaluate:main` — that's the #83 consolidation wave work, not a doc/test cleanup.
- Merging `test_dataset_logic::TestDurationDaysPassthrough` into the `EveryQueryBatch` doctest — the tests are more thorough than a doctest should be; leaving them for a separate "test-surface audit" pass.
- Deprecating the legacy two-column label shape (`is_legacy_two_col` branch) — a runtime warning on old parquets is orthogonal; easier to land once we've verified no producer still writes that shape.

## Test plan

- [x] `uv run pytest --doctest-modules src/every_query/` — all doctests green.
- [x] `uv run pytest -q` — non-slow suite passes locally.
- [ ] CI green on push.

### DIFF
--- a/src/every_query/data/schema.py
+++ b/src/every_query/data/schema.py
@@ -117,10 +117,13 @@ def empty_task_query_df() -> pl.DataFrame:
         >>> df = empty_task_query_df()
         >>> df.height
         0
-        >>> df.columns
-        ['subject_id', 'prediction_time', 'query', 'duration_days', 'boolean_value']
-        >>> df.dtypes
-        [Int64, Datetime(time_unit='us', time_zone=None), String, Float32, Boolean]
+        >>> for name, dtype in df.schema.items():
+        ...     print(f"{name}: {dtype}")
+        subject_id: Int64
+        prediction_time: Datetime(time_unit='us', time_zone=None)
+        query: String
+        duration_days: Float32
+        boolean_value: Boolean
     """
     return pl.DataFrame(
         schema={
--- a/src/every_query/data/schema.py
+++ b/src/every_query/data/schema.py
@@ -75,10 +75,8 @@ class TaskQuerySchema(LabelSchema):
         ...      "query": "ICD//I10", "duration_days": 30.0, "boolean_value": False},
         ... ])
         >>> aligned = TaskQuerySchema.align(data)
-        >>> set(f.name for f in aligned.schema) >= {
-        ...     "subject_id", "prediction_time", "query", "duration_days", "boolean_value"
-        ... }
-        True
+        >>> [f.name for f in aligned.schema]
+        ['subject_id', 'prediction_time', 'boolean_value', 'query', 'duration_days']
 
         Fractional durations are supported:
 
@@ -119,14 +117,10 @@ def empty_task_query_df() -> pl.DataFrame:
         >>> df = empty_task_query_df()
         >>> df.height
         0
-        >>> set(df.columns) == {
-        ...     TaskQuerySchema.subject_id_name,
-        ...     TaskQuerySchema.prediction_time_name,
-        ...     TaskQuerySchema.query_name,
-        ...     TaskQuerySchema.duration_days_name,
-        ...     TaskQuerySchema.boolean_value_name,
-        ... }
-        True
+        >>> df.columns
+        ['subject_id', 'prediction_time', 'query', 'duration_days', 'boolean_value']
+        >>> df.dtypes
+        [Int64, Datetime(time_unit='us', time_zone=None), String, Float32, Boolean]
     """
     return pl.DataFrame(
         schema={
--- a/src/every_query/predict/README.md
+++ b/src/every_query/predict/README.md
@@ -11,7 +11,8 @@ predict/
 ├── predict.py            → EQ_predict (inference-only Hydra main)
 ├── schema.py             → PredictionSchema (TaskQuerySchema + censor_prob + occurs_prob)
 ├── configs/
-│   └── predict.yaml      → required: model_run_dir, tasks_parquet, output_parquet
+│   └── predict.yaml      → required: model_run_dir, tasks_dir, output_parquet
+│                           optional: ckpt_name, split (held_out|tuning)
 └── external_tasks/       → convert + aggregate tasks outside EQ's native vocabulary
     ├── aces_to_eq.py     → ACES task parquet → EQ task parquet
     ├── process_composite.py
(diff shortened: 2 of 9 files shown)


src/every_query/data/schema.py
```diff
@@ -119,14 +117,10 @@ def empty_task_query_df() -> pl.DataFrame:
...
         0
-        >>> set(df.columns) == {
-        ...     TaskQuerySchema.subject_id_name,
-        ...     TaskQuerySchema.prediction_time_name,
-        ...     TaskQuerySchema.query_name,
-        ...     TaskQuerySchema.duration_days_name,
-        ...     TaskQuerySchema.boolean_value_name,
-        ... }
-        True
+        >>> df.columns
```
mmcdermott: you can just print the schema dictionary, right?

src/every_query/predict/external_tasks/aces_to_eq.py
```diff
@@ -47,7 +47,17 @@ def create_eq_task_df(
 
     joined_df = aces_shard_df.join(eq_shard_all_tasks_df, on=["subject_id", "prediction_time"], how="left")
 
-    assert joined_df.select(pl.col("censored").null_count()).item() == 0
+    # Production-safe validation: ``assert`` gets stripped under ``python -O``, which
```
mmcdermott: you don't need to explain not using assert in a comment.

src/every_query/predict/README.md
mmcdermott: you can make this guaranteed to stay live (at least the directory list) by making this a doctest with https://github.com/mmcdermott/pretty-print-directory (which will be in doctest scope as soon as it is installed, no need to import it) but its up to you. You lose your comments on it, which are nice, so there is a real tradeoff.

src/every_query/data/schema.py
mmcdermott: Tried `>>> df.schema` directly but the one-line `Schema({...})` repr lands at ~155 chars — over ruff's 110-char limit (that's the full `Datetime(time_unit='us', time_zone=None)` repr eating the budget even with 5 short columns).

Went with an iteration instead — still prints name + dtype for every column, one per line, no per-line over-110 issue:

```python
>>> for name, dtype in df.schema.items():
...     print(f"{name}: {dtype}")
subject_id: Int64
prediction_time: Datetime(time_unit='us', time_zone=None)
query: String
duration_days: Float32
boolean_value: Boolean
```

Let me know if you'd rather accept the long-line noqa.

src/every_query/predict/external_tasks/aces_to_eq.py
mmcdermott: Fair — dropped the rationale, kept the factual one-liner about what the check enforces.  Comment now reads "Every (subject_id, prediction_time) in the ACES shard must be present in the EQ all-tasks shard — a null ``censored`` column means the left-join failed to match."

src/every_query/predict/README.md
mmcdermott: Good call — done.  Tree is now a `print_directory` doctest so it stays live:

~~~python
>>> from pretty_print_directory import PrintConfig
>>> print_directory("src/every_query/predict", config=PrintConfig(ignore_regex="__pycache__"))
├── README.md
├── __init__.py
├── configs
│   └── predict.yaml
├── external_tasks
...
~~~

You were right about the trade-off though — the per-entry inline annotations (`→ EQ_predict...`, `→ PredictionSchema (TaskQuerySchema + censor_prob + occurs_prob)`) are gone from the tree itself.  Split those into a "Key files" prose list right under the tree so the API hints survive without blocking the auto-verification.  One extra section, but the tree can't lie anymore.

tests/training_validity/test_training_validity.py
```diff
@@ -297,98 +301,86 @@ def test_trained_model_learns_occurs_and_censor(
...
+            "-m",
+            "every_query.evaluate.evaluate",
+            f"predictions_parquet={predictions_parquet!s}",
+            f"metrics_parquet={metrics_parquet!s}",
+        ],
+        env=ENSURE_ENV_PLACEHOLDERS,
+        timeout=60.0,
     )
-    for r in per_cell.iter_rows(named=True):
+    metrics = pl.from_arrow(pq.read_table(metrics_parquet)).sort(["query", "duration_days"])
```
mmcdermott: for reading, with no schema intermediary, why not just pl.read_parquet?

mmcdermott: You're right — no schema alignment on the read side, just unnecessary ceremony.  Fixed in a620d0d — both reads (metrics + predictions) are now `pl.read_parquet(...)` directly, and dropped the now-unused `pyarrow.parquet` import.

PR #141, opened 2026-04-22
Add PerTaskAurocCallback for AUROC monitoring and evaluation

gkondas:
Adds PerTaskAurocCallback that logs per-(query, duration) AUROC against a pre-built tracking task set every N steps, plus EQ_generate_tracking_tasks for generating the labels and supporting eval sweep scripts.


gkondas (commented):
**Clarify tracking task semantics (zip vs cross-join)**

The PR uses a **cross-join** shape (`subjects × K prediction_times × N sampled tuples`) which is actually mini-evaluation semantics, not pretraining semantics — but the module docstring frames it as "sibling to `sample_tasks`" which suggests a pretraining-style zip. Worth making the intent explicit because the two choices have very different meanings:

- **Zip (pretraining shape)**: `N` tasks each get their own `M` iid contexts → `N*M` rows total, distributionally matches `tuning/monitor` and the training loss. But with `N=50, M=50` each task gets ~50 labels, many censored or single-class, so most per-task AUROCs will be null/noisy.
- **Cross-join (current PR, mini-eval shape)**: every subject scored on every sampled task → thousands of labels per task, stable AUROC, apples-to-apples across tasks, matches what `EQ_evaluate` eventually reports. But the distribution differs from training.

For monitoring per-task AUROC the cross-join is the right call (statistical power wins), but the framing should own that. Concretely:

1. Update the module docstring to say this is a **sampled-task subset of evaluation shape**, not a variant of pretraining shape.
2. Consider renaming `build_paired_index_df` → something like `build_sampled_eval_index_df` and cross-referencing `build_evaluation_index_df` in `sample_evaluation_tasks.py` as the dense analogue.
3. Optional: a one-line comment in `sample_tracking_tasks_config.yaml` noting "distribution ≠ training distribution; this measures eval-shape per-task trajectories" so future-you doesn't wonder why `tuning/monitor` and `tuning/per_task/occurs_auroc_mean` disagree.

gkondas:
Closing during branch cleanup (returning after ~1mo). Work preserved on the PR; branch tip 36f47cf captured if we want to revive per-task AUROC monitoring later.

PR #151, opened 2026-04-22
add code_sample_frac option to EQ_generate_training_tasks

gkondas:
Lets a sweep restrict the query-code pool to a random fraction of the full codes.parquet vocabulary, seeded deterministically so every worker agrees on the same subset. The chosen subset is persisted once to {out_dir}/sampled_codes.parquet for post-hoc audit of which codes the run saw.

### DIFF
--- a/src/every_query/generate_tasks/configs/sample_tasks_config.yaml
+++ b/src/every_query/generate_tasks/configs/sample_tasks_config.yaml
@@ -33,6 +33,13 @@ duration_max: 731
 # Minimum prior events a subject must have accumulated before a prediction time is valid.
 min_context_per_subject: 50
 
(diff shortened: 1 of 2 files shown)


gkondas:
from chatgpt: This mostly looks good, but I think sampled_codes.parquet can become stale relative to the actual worker outputs. Right now we recompute the sampled vocab from the current seed / code_sample_frac, but only rewrite the sidecar if it doesn’t exist or overwrite=true. If someone reuses an existing out_dir with a different seed or fraction, newly generated label shards could be based on a different sampled vocab than the one recorded in sampled_codes.parquet. Can we either make the sidecar filename config-specific, or validate that an existing sampled_codes.parquet matches the current sampled set and fail otherwise?

PR #157, opened 2026-04-23
Remove train-code manual selection from training config

gkondas:
## Summary
- Drop the `query.codes` / `train_codes` compose-group knob from the training config; the queried-code set now comes from the task-labels parquet written by `EQ_generate_training_tasks` (upstream preprocessing).
- Simplify `wandb` run naming and drop the `num_query_codes`/`query_codes` wandb config fields.
- Extend `resume_check` with a `LEGACY_REMOVED_KEYS` allow-list (starting with `query`) so run dirs started before this change stay resumable.
- Add `tests/test_resume_check.py` covering the legacy-key stripping path; trim now-obsolete assertions in `tests/test_cli_smoke.py`.

## Test plan
- [ ] \`uv run pre-commit run --all-files\` (green locally)
- [ ] \`uv run pytest tests/test_resume_check.py tests/test_cli_smoke.py -v\`
- [ ] Full test suite in CI
- [ ] Spot-check: resume a pre-#64 run dir and confirm no \`query\` drift error
- [ ] Spot-check: fresh \`EQ_train\` run succeeds and W&B run name uses the new \`\${run_id}\` form

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -181,12 +181,6 @@ cython_debug/
 #  option (not recommended) you can uncomment the following to ignore the entire idea folder.
 #.idea/
 
-src/every_query/evaluate/conf/eval_codes/
-src/every_query/predict/external_tasks/query_codes/
-
-# Generated train code YAMLs
-src/every_query/train_codes/
-
 src/every_query/evaluate/outputs/
 
 # AI
--- a/README.md
+++ b/README.md
@@ -211,21 +211,6 @@ env vars when CLI config values are supplied).
 gated vars to `${oc.env:VAR,???}` / `${oc.env:VAR,default}` form (Hydra-native required
 or optional-with-fallback) and eventually retire `ensure_env()` entirely.
 
-### Known gotcha: code-group YAMLs
-
-`train/configs/config.yaml`, `evaluate/conf/eval_config.yaml`, and
-`evaluate/conf/gen_tasks_config.yaml` all pull a default `{train,eval}_codes/<hash>.yaml`
-that is (a) generated out-of-band and (b) explicitly `.gitignore`d — so a fresh clone
-can't compose them. Workaround until [#64](https://github.com/payalchandak/EveryQuery/issues/64)
-lands:
-
-- Pass `--config-dir=/path/to/your/codes_dir code_group_name=...`, or
-- Generate them locally via
-    `python -m every_query.paper_experiments.sample_codes.sample_train_codes` (note: currently
-    has a hardcoded MIMIC path — #85 will parameterize it).
-
-The smoke-test fixture in `tests/test_cli_smoke.py` shows the minimal shape of each file.
-
 ## Development
 
 ```bash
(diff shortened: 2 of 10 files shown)


gkondas:
Code review findings:

- P3: `README.md` still has stale fresh-clone guidance under "Known gotcha: code-group YAMLs". It says `train/configs/config.yaml` pulls a gitignored `train_codes/<hash>.yaml` and points readers to the deleted smoke fixture shape, but this PR removes that train-code default.

- P3: `src/every_query/train/configs/_demo_train.yaml` still defines `query.codes`, even though training no longer has a query-codes knob and `train.py` no longer reads it. It looks harmless at runtime, but keeps a dead override surface in the demo config and contradicts the updated train README.

Verification: `uv run pytest tests/test_resume_check.py tests/test_cli_smoke.py -q` passed (`8 passed`, warnings only).

gkondas:
Reviewed locally by checking out the branch and grepping for stragglers. LGTM.

**Verified removed (no remaining refs):**
- `query.codes`, `num_query_codes`, and the `train_codes` compose group are gone from `src/every_query/train/` and `tests/`
- `train.py` no longer imports `ListConfig` and no longer has the `isinstance(cfg.query.codes, ListConfig)` guard
- `config.yaml` no longer lists `train_codes` in `defaults:`; W&B `name` / `config` no longer reference `${train_codes.codes}`
- `tests/test_cli_smoke.py` drops the `smoke_config_dir` fixture and the per-entrypoint `extra_args` parametrization
- `tests/training_validity/test_training_validity.py` no longer passes `query.codes=[...]`
- `resume_check.py`'s `LEGACY_REMOVED_KEYS = {"query"}` is correctly stripped from both `old_cfg` and `new_cfg` before diffing, and the new `tests/test_resume_check.py` covers both the legacy-strip path and a real drift-rejection case

**Legitimately kept (out of scope for this PR):**
- `src/every_query/generate_tasks/sample_tasks.py` / `sample_evaluation_tasks.py` — `query_codes`/`read_query_codes` are the upstream preprocessing that now *produces* the task-labels parquet. This is exactly the source the PR description says the training query set now comes from.
- `src/every_query/paper_experiments/sample_codes/` — independent scripts for sampling codes from MEDS into YAMLs; not consumed by `EQ_train` anymore but still used by the paper-experiments pipeline.
- `src/every_query/predict/external_tasks/configs/aces_to_eq.yaml` — still has `- query_codes: admin_codes` as a compose-group default. This is predict-side, not train-side.

**One minor thing worth flagging:** the `.gitignore` diff drops `src/every_query/predict/external_tasks/query_codes/`, but `aces_to_eq.yaml` still declares `query_codes: admin_codes` as a required compose-group default (and there's no `admin_codes.yaml` in the repo). If that directory is still expected to be generated out-of-band on the predict side, removing the ignore line is probably premature — it just means future generated YAMLs there will show up as untracked. Since this PR is about the training config, might be worth either (a) leaving that `.gitignore` line in, or (b) calling out in the description that you're intentionally punting predict-side cleanup to a follow-up.

Otherwise clean — happy with the `LEGACY_REMOVED_KEYS` approach for keeping old run dirs resumable.

PR #159, opened 2026-04-23
fix(config): use fully-qualified target for trainer callbacks

### LINKED ISSUE #158: trainer.callbacks._target_ hardcodes __main__ and breaks the EQ_train entry point
## Problem

`src/every_query/train/configs/config.yaml` sets

```yaml
trainer:
  callbacks:
    _target_: __main__.values_as_list
```

This only resolves when the training module itself is `__main__` (i.e. `python -m every_query.train.train`). When launched via the installed console script `EQ_train` (entry point `every_query.train.train:main`), `__main__` is the entry-point stub, so Hydra fails to import `__main__.values_as_list`:

```
ModuleNotFoundError: No module named '__main__.values_as_list'; '__main__' is not a package
...
hydra.errors.InstantiationException: Error locating target '__main__.values_as_list'
full_key: trainer.callbacks
```

## Reproduction

```
uv run EQ_train   # any invocation that doesn't use `python -m`
```

## Suggested fix

Point `_target_` at the real, stable path:

```yaml
trainer:
  callbacks:
    _target_: every_query.train.train.values_as_list
```

That works under both `EQ_train` and `python -m every_query.train.train`.

## Workaround

Pass the override at the CLI until fixed:

```
uv run EQ_train trainer.callbacks._target_=every_query.train.train.values_as_list ...
```

Environment: everyquery 0.3.1, Python 3.13, Linux.

gkondas:
## Summary
- Replace `__main__.values_as_list` with `every_query.train.train.values_as_list` in `src/every_query/train/configs/config.yaml` so Hydra can resolve the target under the installed `EQ_train` console script (not just `python -m every_query.train.train`).
- Matches the pattern already used in `_demo_train.yaml`.

Fixes #158.

## Test plan
- [ ] `uv run EQ_train --help` succeeds (no `InstantiationException` for `trainer.callbacks`)
- [ ] `uv run python -m every_query.train.train --help` still works (regression check)
- [ ] Short cluster-side `EQ_train` run confirms training starts

### DIFF
--- a/src/every_query/train/configs/config.yaml
+++ b/src/every_query/train/configs/config.yaml
@@ -59,7 +59,7 @@ trainer:
     prefix: "" # a string to put at the beginning of metric keys
     entity: ${oc.env:WANDB_ENTITY}
   callbacks:
-    _target_: __main__.values_as_list
+    _target_: every_query.train.train.values_as_list
     learning_rate_monitor:
       _target_: lightning.pytorch.callbacks.LearningRateMonitor
     model_checkpoint:


gkondas:
Review assessment: no blocking issues found.

I checked the one-line config change and it looks correct. `EQ_train` is packaged from `every_query.train.train:main`, so `__main__.values_as_list` is brittle under the installed console script, while `every_query.train.train.values_as_list` is import-stable in both invocation paths. This also brings `config.yaml` in line with `_demo_train.yaml`.

I also verified locally that both of these succeed after the change:
- `uv run EQ_train --help`
- `uv run python -m every_query.train.train --help`

Residual gap: I did not run a real cluster-side training job, so the only unverified part of the PR test plan is the end-to-end start on cluster infra. Outside of that, this looks ready to merge.

gkondas:
Reported in #158 (with repro, root cause, and test-coverage gap). This PR looks like the right fix — thanks for picking it up so fast.

One small suggestion: consider also adding a regression test that runs `EQ_train` end-to-end against the default `config.yaml` (the combo that currently has no coverage — `test_cli_smoke.py` stops at `--help`, and `test_train_cli.py` uses `python -m ...` + `_demo_train.yaml`, both of which independently dodge the bug).

PR #161, opened 2026-04-24
feat: sample eval-task codes to gitignored YAML

gkondas:
## Summary
- Teaches `read_query_codes` to load codes from a YAML file (list form, or mapping with a `codes:` key) so a pre-sampled, gitignored code universe can flow through the normal config path
- Points `sample_evaluation_tasks_config.yaml` at the new YAML-backed code source and updates the evaluation duration grid to `[30, 90, 180, 365, 731]`
- Updates `.gitignore` to exclude `src/every_query/paper_experiments/sample_codes/outputs/` where the sampled YAML is written

Note: the sampling script itself (`sample_eval_task_codes.py`) was removed from this PR and now lives in an external experimental repo. This PR is just the plumbing that lets the evaluator consume its YAML output.

## Test plan
- [ ] Generate a codes YAML externally and confirm `read_query_codes` returns a deterministic, deduped list
- [ ] Confirm the output directory is gitignored
- [ ] Run `sample_evaluation_tasks` end-to-end against the YAML-backed config


gkondas:
Codex review:

The design generally makes sense. Letting `read_query_codes` accept a YAML file with a `codes:` list is a reasonable extension of the existing `codes` contract: it keeps large sampled code sets out of git while still letting `sample_evaluation_tasks` consume a reproducible, explicit code universe through the normal config path.

A few things I would tighten before treating this as ready:

- Please add a small regression test for `read_query_codes` with a YAML file, including duplicate codes, so this path is covered alongside the existing list/parquet behavior. Codecov is also flagging this new branch as essentially untested.
- The PR summary says this adds a script to sample eval-task codes, but the diff only changes `.gitignore`, the evaluation config, and `read_query_codes`. If the script was intentionally already present, the summary should say this PR wires generated YAML into the evaluator; otherwise the script/output-path change seems missing.
- The duration-grid change from `[1, 7, 30, 90, 365]` to `[30, 90, 180, 365, 731]` is a substantive evaluation-design change and is unrelated to YAML-backed code sampling. It may be the right paper-experiment grid, but it should be called out explicitly in the PR body or split from the plumbing change.
- Minor robustness: the YAML branch assumes the file is either a list or a mapping with `codes`. That is fine for internal generated files, but a clear error for malformed YAML or missing `codes` would make CLI failures easier to diagnose.

Net: the addition is reasonable and the direction fits the repo. I would not block on the design, but I would want the test and PR-scope clarification before merging.

gkondas:
Updated the PR summary to clarify scope: the sampling script itself was removed and now lives in an external experimental repo, so this PR is just the plumbing that wires YAML-backed codes into the evaluator. Also called out the duration-grid change explicitly in the summary.

PR #162, opened 2026-04-24
feat: add occurs_loss_weight hyperparameter

payalchandak:
## Summary

Analysis in #139 showed that censor and occurs losses operate at different scales during training, and that early stopping on the combined `tuning/loss` can be dominated by the censor component. Matt suggested downweighting censor loss significantly during training to improve occurs prediction quality.

This PR adds a single mixing parameter `occurs_loss_weight` (float in [0,1], default 0.5) to `EveryQueryModel`. The training loss becomes:

```
loss = occurs_loss_weight * occurs_loss + (1 - occurs_loss_weight) * censor_loss
```

- **Default (0.5)**: both losses receive equal weight (backward compatible behavior)
- **occurs_loss_weight > 0.5**: upweights occurs loss relative to censor loss
- Example: `occurs_loss_weight=0.91` gives ~10x occurs vs censor weighting

### Changes

- `src/every_query/model/model.py`: added `occurs_loss_weight` parameter to `EveryQueryModel.__init__`, stored on the model, included in `hparams`, and used in `_forward` loss computation
- `src/every_query/train/configs/config.yaml`: added `occurs_loss_weight: 0.5` to model config block
- `tests/test_training.py`: updated `test_loss_decomposes` to use the weighted formula

### DIFF
--- a/src/every_query/model/model.py
+++ b/src/every_query/model/model.py
@@ -281,6 +281,7 @@ class EveryQueryModel(torch.nn.Module):
         model_name: str = "answerdotai/ModernBERT-base",
         num_hidden_layers: int | None = None,
         config_overrides: dict[str, Any] | None = None,
+        occurs_loss_weight: float = 0.5,
     ):
         super().__init__()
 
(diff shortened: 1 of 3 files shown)


gkondas (approved):
Looks good

PR #172, opened 2026-04-28
Stream predictions to parquet instead of buffering in memory

payalchandak:
## Summary
Refactored the prediction pipeline to stream per-batch results directly to parquet instead of accumulating all predictions in memory. This eliminates memory scaling with cohort size by writing each batch as a row group immediately after prediction.

## Key Changes
- **Removed `_gather_probabilities()`**: Eliminated the function that accumulated all per-batch predictions in memory before writing to disk
- **Added `_StreamingPredictionWriter` class**: New Lightning `BasePredictionWriter` callback that:
  - Opens a parquet writer in `setup()` with the canonical schema
  - Appends one row group per batch in `write_on_batch_end()`
  - Provides idempotent `close()` for safe cleanup from both Lightning's teardown and main's finally block
  - Handles empty cohorts correctly by pre-building schema from zero-row aligned table
  - Tracks offset into `schema_df` to extract identifiers for each batch
- **Updated `main()` function**:
  - Instantiates `_StreamingPredictionWriter` and registers it as a trainer callback
  - Calls `trainer.predict()` with `return_predictions=False` so Lightning drops each batch after the writer processes it
  - Wraps prediction in try/finally to ensure writer is closed even on exceptions
  - Validates row count only on success path (partial writes are legitimate on exception paths)
  - Removed in-memory accumulation and post-hoc alignment logic

## Implementation Details
- Partial-write semantics: Python-level exceptions mid-stream still produce valid parquet with completed batches; hard kills (SIGKILL) cannot be recovered
- Correctness depends on dataloader iterating in `schema_df` row order, enforced by existing `SequentialSampler` guard
- Per-batch Float32 casting mirrors original behavior to avoid repeated f64→f32 coercion
- Added imports: `pyarrow` and `lightning.pytorch.callbacks.BasePredictionWriter`

https://claude.ai/code/session_017xxwNdPR1mzo3oZ2v4hpKf

### DIFF
--- a/src/every_query/predict/predict.py
+++ b/src/every_query/predict/predict.py
@@ -393,9 +393,7 @@ def main(cfg: DictConfig) -> None:
     # class's ``MisconfigurationException``.  The SequentialSampler check above
     # guarantees order preservation.
     try:
-        trainer.predict(
-            model=model, dataloaders=dataloader, ckpt_path=None, return_predictions=False
-        )
+        trainer.predict(model=model, dataloaders=dataloader, ckpt_path=None, return_predictions=False)
         # Success-path invariant — MTD guarantees ``schema_df`` preserves the input
         # labels frame's length + order, so the writer should have streamed one
         # prediction per schema_df row.  A mismatch is a silent invariant violation;
--- a/src/every_query/predict/predict.py
+++ b/src/every_query/predict/predict.py
@@ -356,6 +356,23 @@ def main(cfg: DictConfig) -> None:
     logger.info(f"Loading tasks from {tasks_dir} (split={split})")
 
     train_cfg, model, trainer = setup_model(model_run_dir, ckpt_name=ckpt_name)
+
+    # ``_StreamingPredictionWriter`` opens a single ``output_parquet`` and tracks a
+    # process-local offset into ``schema_df``.  Under multi-process predict (DDP /
+    # multi-device / multi-node), every rank would open the same path and start slicing
+    # ``schema_df`` from offset 0 — corrupting the parquet and mispairing probabilities
+    # with identifiers.  ``setup_model`` instantiates the trainer from the saved training
+    # config, so a model trained with ``trainer.devices > 1`` carries that strategy into
+    # predict.  Refuse loudly; predict is single-pass by design (see module docstring).
+    if trainer.world_size > 1:
+        raise RuntimeError(
+            f"EQ_predict requires single-process prediction, but the trainer instantiated from "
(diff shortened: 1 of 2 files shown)


gkondas:
**Comment from Codex**

I reviewed the diff for PR #172. The streaming direction looks sound for the current single-process prediction path: `predict_step` already returns CPU tensors, the existing `SequentialSampler` guard still protects row order, per-batch `PredictionSchema.align(...)` preserves the canonical output schema, and the empty-cohort schema setup is a good replacement for the old `_gather_probabilities([])` behavior.

One risk I would address before merging:

- `src/every_query/predict/predict.py:274` / `src/every_query/predict/predict.py:297`: `_StreamingPredictionWriter` opens the final `output_parquet` inside the Lightning callback and tracks identifiers with a process-local `_offset`. Lightning callbacks run in each predict process under multi-device/DDP-style strategies, while `setup_model()` instantiates the trainer from the saved training config. If a run was trained/configured with `devices > 1` or a distributed strategy, each rank can open/write the same parquet path, and each rank starts slicing `schema_df` from offset 0. That can corrupt the parquet or pair probabilities with the wrong identifiers. Suggested fix: either force/validate single-process prediction before attaching this writer, or write rank-local shards keyed by rank/global batch indices and merge on rank zero.

Test notes: CI is green and Codecov reports patch coverage at 91.11%. The existing subprocess tests cover the CPU single-device happy path and row-order preservation. I would add a small writer-level test for one batch plus the empty cohort path, and either a guard test that distributed/multi-device trainer configs are rejected or a distributed/sharded write test if multi-device prediction should be supported.

gkondas:
**Comment from Codex**

I reviewed the current PR head (`a963ab8`) after the follow-up changes. I do not see any blocking issues in this revision.

The earlier concern about multi-process prediction corrupting a shared parquet writer appears addressed by `_check_single_process_trainer(...)`, and the writer now streams to `.tmp` files before replacing the canonical outputs, which is a better failure mode than exposing a partially written final parquet. The new `tests/test_predict_logic.py` coverage hits the important internal cases: single batch, multiple batches with offset progression, empty cohort output, and the multi-process guard.

One non-blocking note: the final output replacement is still atomic per file rather than atomic across the predictions, optional embeddings, and timing files as a set. That is a pragmatic improvement over direct writes, but a failure during the rename phase could theoretically leave a mixed set of canonical artifacts. I would not block on that unless this CLI needs transactional multi-file output semantics.

Check status when reviewed: `Code Quality PR` was passing, and the `Tests` workflow for this head was still in progress.

PR #173, opened 2026-04-28
Add save_embeddings flag to write pooled query embeddings

payalchandak:
## Summary
Adds a new `save_embeddings` configuration flag to the EQ_predict CLI that enables writing per-row pooled query embeddings to a sibling parquet file alongside the predictions output.

## Key Changes
- **New `save_embeddings` config flag** in `predict.yaml`: Optional boolean (default `false`) that controls whether embeddings are written to a sibling parquet file
- **`_gather_embeddings()` function**: Concatenates per-batch `query_embed` tensors from Lightning's `predict_step` and returns a polars DataFrame with a fixed-size-list column backed by PyArrow, preserving the fixed-width invariant through parquet serialization
- **`_derive_embeddings_path()` function**: Derives the embeddings sibling path by inserting `.embeddings` before the file suffix (e.g., `predictions.parquet` → `predictions.embeddings.parquet`)
- **Embeddings file protection**: The embeddings sibling participates in the existing `overwrite` guard — attempting to re-run without `overwrite=true` raises `FileExistsError` if the embeddings file already exists
- **Output format**: The embeddings parquet contains TaskQuerySchema identifier columns (subject_id, prediction_time, query, duration_days) plus an `embedding` column of type `fixed_size_list<float32>[hidden_size]`
- **Comprehensive test coverage**: New `test_eq_predict_save_embeddings()` test validates file creation, schema correctness, row alignment with predictions, and clobber-protection behavior

## Implementation Details
- Uses PyArrow's `FixedSizeListArray.from_arrays()` to construct the fixed-size embedding column, ensuring the width invariant is preserved through parquet writes
- Embeddings are row-aligned with predictions via the shared `identifiers` frame source
- The embeddings file is only written if `save_embeddings=true`; the predictions output is unaffected by this flag

https://claude.ai/code/session_01H9uM1WaMuZxgwd2qkKtwjr

### DIFF
--- a/src/every_query/predict/configs/predict.yaml
+++ b/src/every_query/predict/configs/predict.yaml
@@ -39,6 +39,15 @@ split: held_out
 # same output path without this flag raises ``FileExistsError`` at startup.
 overwrite: false
 
+# Optional: also write per-row pooled query embeddings to a sibling parquet at
+# ``<output_parquet stem>.embeddings.parquet`` (e.g. ``predictions.parquet`` →
+# ``predictions.embeddings.parquet``).  The sibling carries the same TaskQuerySchema
+# identifiers as the predictions file plus a single ``embedding`` column of
+# ``fixed_size_list<float32>[hidden_size]``.  Default false.  Subject to the same
+# ``overwrite`` guard as ``output_parquet`` — re-running without ``overwrite=true``
+# raises ``FileExistsError`` at startup if the sibling already exists.
+save_embeddings: false
+
 hydra:
   output_subdir: null
   job:
--- a/src/every_query/predict/predict.py
+++ b/src/every_query/predict/predict.py
@@ -430,15 +430,13 @@ def main(cfg: DictConfig) -> None:
     logger.info(f"Wrote {out.height} predictions to {output_parquet}")
 
     if embeddings_output_parquet is not None:
-        # Prefer the runtime tensor shape (the actual data we just produced) and fall
-        # back to the model's config when there are zero batches — keeps the sidecar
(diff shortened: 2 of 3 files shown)


gkondas:
**Comment from Codex**

I reviewed PR #173. The feature shape is reasonable: `save_embeddings` is opt-in, the sidecar path is deterministic, the sibling participates in the overwrite guard, and the integration test checks row alignment plus the fixed-size-list parquet type on the normal non-empty path. Code Quality and Tests are both green on `b9608ca`, and Codecov reports all modified coverable lines covered.

One edge case I would tighten before merging:

- `src/every_query/predict/predict.py:264` / `src/every_query/predict/predict.py:429`: `_gather_embeddings([])` returns a Polars `List(Float32)` column, so an empty cohort with `save_embeddings=true` writes an `embedding` column as a variable-length `list<float32>`, not the documented `fixed_size_list<float32>[hidden_size]`. The predictions path already intentionally supports empty outputs via `_gather_probabilities([])`, and the new config comment promises a fixed-size-list sidecar, so this is a schema inconsistency for the empty-input case. Suggested fix: build the empty embedding table from the model/config hidden size, or explicitly reject `save_embeddings=true` when no prediction batches are produced. A small test with an empty tasks parquet would catch this.

Other than that, the non-empty path looks aligned with the existing prediction ordering assumptions: embeddings and probabilities are gathered from the same `pred_batches`, identifiers come from the same `_identifiers_from_schema_df(dataset.schema_df)` source, and the new subprocess test covers the sidecar path end to end.

gkondas (approved):

PR #175, opened 2026-04-28
Add prevalence metric to evaluation output

payalchandak:
## Summary
This PR adds a `prevalence` metric to the evaluation pipeline, computing the positive rate among labeled (non-censored) rows for each query-duration group.

## Key Changes
- **New metric column**: Added `prevalence` to `compute_metrics()` output, calculated as `n_positive / n_occurs_labeled`
  - Returns `null` when `n_occurs_labeled == 0` (no labeled rows in group)
  - Denominator matches `occurs_auroc`, ensuring both metrics are computed over the same cohort
  
- **Documentation updates**:
  - Updated `_metrics_for_group()` docstring to clarify that `prevalence` is not computed at the group level but derived in `compute_metrics()`
  - Updated `compute_metrics()` docstring with detailed explanation of prevalence definition and its relationship to `occurs_auroc`
  - Updated example output to show prevalence values (0.5 for mixed labels, 1.0 for all positive)
  - Updated README diagram to include `prevalence` in output schema
  - Updated empty dataframe shape assertion from 7 to 8 columns

- **Implementation**:
  - Prevalence is derived using Polars conditional expression after assembling the metrics dataframe
  - Explicitly cast to `Float64` for schema stability (consistent with AUROC columns)
  - Prevents drift between prevalence and its definition by computing from the same `n_positive` and `n_occurs_labeled` counts

- **Tests**: Added assertions validating prevalence values (0.5 and 1.0) for the two test query groups

https://claude.ai/code/session_0175oiu2QVBY2rMzP91NJ8DE

### DIFF
--- a/README.md
+++ b/README.md
@@ -194,7 +194,7 @@ EQ_evaluate \
 	metrics_parquet="$OUTPUT_DIR/metrics.parquet"
 ```
 
-Per-`(query, duration_days)` metrics from the predictions parquet — `n_rows`, `n_occurs_labeled`, `n_positive`, `occurs_auroc` (on non-censored rows), `censor_auroc`. See [`evaluate/README.md`](src/every_query/evaluate/README.md).
+Per-`(query, duration_days)` metrics from the predictions parquet — `n_rows`, `n_occurs_labeled`, `n_positive`, `prevalence`, `occurs_auroc` (on non-censored rows), `censor_auroc`. See [`evaluate/README.md`](src/every_query/evaluate/README.md).
 
 ## Configuration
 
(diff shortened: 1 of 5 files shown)


gkondas:
**Codex review**

I reviewed the PR diff for `Add prevalence metric to evaluation output`.

**Findings**

- Non-blocking doc consistency issue: the new `prevalence` column is documented in `src/every_query/evaluate/README.md` and `compute_metrics()`, but two other user-facing schema descriptions still omit it:
  - `README.md`, in the `EQ_evaluate` section, still lists only `n_rows`, `n_occurs_labeled`, `n_positive`, `occurs_auroc`, and `censor_auroc` for the metrics parquet.
  - `src/every_query/evaluate/configs/evaluate.yaml`, the `metrics_parquet` comment still says the output columns are `n_rows, n_occurs_labeled, n_positive, occurs_auroc, censor_auroc`.

No blocking implementation issues found. The `prevalence` expression is derived from the same `n_positive` / `n_occurs_labeled` counts used for the rest of the metrics, handles `n_occurs_labeled == 0` as null, and explicitly casts to `Float64`, which is consistent with the existing schema-stability pattern for AUROC columns.

**Test coverage**

The added E2E assertion covers the normal mixed-label and all-positive labeled cases. One useful follow-up test would be an all-censored group to pin the `prevalence is null` behavior, but I would not block this PR on it. I did not run the PR branch locally; Codecov reports all tests successful for head `ba1cc7a`.

PR #177, opened 2026-04-28
Add query code filtering for training tasks

gkondas:
## Summary
- Replace `codes_dir` config with a more flexible `query_codes` option in `sample_tasks_config.yaml` — accepts an explicit Hydra list, a YAML file (flat list or `{codes: [...]}`), or `null` to default to `\$PROCESSED/metadata/codes.parquet`.
- Move `read_query_codes` resolution out of `run_worker` and into `main`, so workers receive a resolved code list directly.
- Update tests and README to cover the new resolution paths and YAML-file input.

## Test plan
- [ ] `uv run pytest tests/test_sample_tasks.py`
- [ ] `uv run pytest tests/test_generate_tasks.py`
- [ ] Spot-check a Hydra run with `query_codes=[HR,TEMP]` and with a YAML file path
- [ ] Spot-check the `query_codes: null` fallback against `\$PROCESSED/metadata/codes.parquet`

### DIFF
--- a/README.md
+++ b/README.md
@@ -134,6 +134,22 @@ Sweep across shards with
 `python -m every_query.generate_tasks.sample_tasks -m input_shard=0,1,2,… task_shard=range(0,K)`.
 Each worker writes labeled task parquets under `$TASK_DIR/{split}/*.parquet` idempotently. Output columns conform to [`TaskQuerySchema`](src/every_query/data/schema.py) — `subject_id, prediction_time, query, duration_days, boolean_value` — where `boolean_value` is a nullable three-valued label (`null` = censored, `True` = event occurred in `[prediction_time, prediction_time + duration_days)`, `False` = observed-but-no-event).
 
+`query_codes=` is optional for training. Leave it unset/null to sample query codes from
+`$PROCESSED/metadata/codes.parquet`, or set it to an inline list / YAML path to restrict which
+codes can be sampled as queries. YAML files may be a flat list or a mapping with a `codes:`
+key. This does not remove codes from patient histories.
+
+```bash
+EQ_generate_training_tasks query_codes=/path/to/train_query_codes.yaml …
+```
+
+```yaml
+# train_query_codes.yaml
+codes:
+  - HR
+  - TEMP
+```
+
 ### 2b. Generate evaluation task labels
 
 ```bash
--- a/conftest.py
+++ b/conftest.py
@@ -182,8 +182,8 @@ def eq_preprocessed_dataset(simple_static_MEDS: Path, tmp_path_factory: pytest.T
     """Runs ``EQ_process_data do_demo=True`` against ``simple_static_MEDS``.
 
     Returns the tensorized-cohort output dir (``final/``), ready to feed into
-    ``EQ_generate_training_tasks`` and ``EQ_train`` as ``codes_dir`` and ``tensorized_cohort_dir``
-    respectively.
+    ``EQ_generate_training_tasks`` and ``EQ_train`` as ``$PROCESSED`` and
+    ``tensorized_cohort_dir`` respectively.
 
     Sibling ``intermediate/`` dir is also produced (needed as ``data_dir`` for
     ``EQ_generate_training_tasks``) and stored alongside ``final/`` — callers can reach it via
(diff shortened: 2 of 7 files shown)


gkondas:
## Code Review

### Overview
Replaces the `codes_dir` path-config knob with a more flexible `query_codes` option in `sample_tasks_config.yaml`, accepting:
- `null` → falls back to `$PROCESSED/metadata/codes.parquet`
- a Hydra inline list (`query_codes=[HR,TEMP]`)
- a YAML file path (flat list or `{codes: [...]}`)

Resolution moves from `run_worker` up into `main`, so workers receive a pre-resolved code list. Tests cover the new YAML/env-fallback paths; README + module README updated.

### Strengths
- **Cleaner separation of concerns**: `run_worker` is now a pure function over a code list, not a directory — easier to test and reason about.
- **Solid test coverage** for the new resolution branches: env fallback present/absent, flat-list YAML, `codes:`-keyed YAML, mapping-without-codes rejection, non-list rejection, inline dedup.
- **Order-preserving dedup** on YAML-loaded lists matches the inline-list branch — good consistency.
- README + module README both updated with examples; conftest/tests migrated cleanly.

### Issues / Suggestions

**1. Backward-incompatibility for callers using `codes_dir=`** (medium)
Hydra configs that previously passed `codes_dir=/path` (e.g. SLURM scripts, downstream tooling, prior sweep configs) will now fail with a struct-mode error since `codes_dir` is removed from the schema. Worth either:
- Mentioning the breaking change explicitly in the PR description / a changelog note, or
- Grepping the repo (sweep configs, scripts/, docs) for stale `codes_dir=` references.

**2. Error message points only to env vars, not `.env`** (small)
`sample_tasks.py:421-424` says "export `$PROCESSED`", but the repo convention (per `main`'s own docstring at line 581) is that `$PROCESSED` lives in a gitignored `.env` file loaded via python-dotenv. Suggest:
> "...pass `query_codes=...` or set `$PROCESSED` (typically in `.env`) so the sampler can read `$PROCESSED/metadata/codes.parquet`."

**3. YAML-resolution semantics for a dict without `codes` key are now silently different** (small)
Old: `data[\"codes\"] if isinstance(data, dict) else data` → KeyError on `{query_codes: [...]}`.
New: falls through to the list check and raises ValueError with `\"list of codes, got dict\"`.
That's friendlier, and there's a test for it — but the error message could name the expected key (`\"YAML mapping must have a 'codes:' key\"`) to make typos like `query_codes:` (which the test demonstrates) immediately obvious.

**4. Minor: `read_query_codes` is also called by `sample_evaluation_tasks.py`** (informational)
Verified that eval still passes a resolved `codes_dir` path (never `None`), so the new env-fallback branch is training-only in practice. Not a problem — just worth noting that the eval CLI knob remains `codes`/`codes_dir` while training is now `query_codes`. The asymmetry is documented in `generate_tasks/README.md` but might be worth aligning in a follow-up for consistency.

**5. Test naming nit** (tiny)
`test_yaml_code_list_restricts_sampled_queries` ends with `assert set(labels[\"query\"].to_list())` — a non-empty assertion that's implicit from the subset check passing on real samples, but the intent reads cleanly. Fine to keep.

### Correctness / Performance / Security
- No correctness concerns in the resolution logic.
- No perf implications — resolution moves up one call frame, runs once per `main` invocation instead of once per worker entry (slight win, immaterial).
- No security concerns; YAML loaded with `safe_load`.

### Verdict
Looks good to merge once the backward-compat callers (`codes_dir=…`) are confirmed clean. Nice refactor — the worker-purity improvement alone is worth it.

gkondas:
Followup on issue #1 (backward-compat for `codes_dir=` callers): grepped the repo for any lingering references —

```
src/every_query/generate_tasks/configs/sample_evaluation_tasks_config.yaml
src/every_query/generate_tasks/sample_evaluation_tasks.py
tests/test_generate_evaluation_tasks_cli.py
```

All remaining `codes_dir` references are confined to the **evaluation** pipeline, which this PR intentionally leaves alone. No stale training-side callers. Safe to merge from a backward-compat standpoint; the training/eval CLI asymmetry (`query_codes` vs `codes_dir`) is worth a follow-up but isn't a blocker.

PR #179, opened 2026-04-30
Add timing artifact to EQ_predict output

gkondas:
## Summary
- Wraps `trainer.predict` in a `time.perf_counter()` window in `src/every_query/predict/predict.py`.
- Always writes a one-row `<output>.timing.parquet` sidecar with `total_seconds`, `n_rows`, `n_tasks`, `seconds_per_row`, `seconds_per_task` — mirrors the existing embeddings sibling pattern (`_derive_embeddings_path` / overwrite guard).
- Per-task time is reported as a uniform average (`total_seconds / n_tasks`); batches mix tasks under `SequentialSampler`, so true per-task wall time would require a more invasive change. The single-row schema can grow to per-task rows later without renaming the artifact.

## Test plan
- [ ] Run `EQ_predict` on an existing eval task on the cluster and confirm `<output>.timing.parquet` exists.
- [ ] `total_seconds` roughly matches SLURM wall-clock for the predict step.
- [ ] `n_rows` matches predictions parquet row count; `n_tasks` matches `predictions.select([query, duration_days]).unique().height`.
- [ ] Predictions parquet bytes unchanged vs. a pre-change run on the same input (timer adds no perturbation).

### DIFF
--- a/src/every_query/predict/predict.py
+++ b/src/every_query/predict/predict.py
@@ -33,6 +33,7 @@ parquets for them.
 """
 
 import logging
+import time
 from importlib.resources import files
 from pathlib import Path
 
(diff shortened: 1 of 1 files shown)


gkondas:
Reviewed PR #179.

I don't see any blocking issues in the implementation. The timing path derivation mirrors the embeddings sidecar pattern, the new sidecar participates in the same overwrite guard, and the measured window is scoped narrowly to `trainer.predict`, which matches the artifact description in the PR body.

I ran the focused predict CLI suite locally:

```bash
uv run pytest tests/test_predict_cli.py
```

Result: `3 passed` in ~71s.

One non-blocking follow-up I would consider before merge: add an explicit assertion in `tests/test_predict_cli.py` that `<output>.timing.parquet` is created and has the expected one-row schema/values. The current focused suite still passes, but it does not directly lock the new artifact contract.

PR #181, opened 2026-05-01
Drop censored rows and emit unique-prediction-times sibling in EQ_generate_evaluation_tasks

gkondas:
## Summary
- Unconditionally drops censored (null `boolean_value`) rows from the labeled output of `EQ_generate_evaluation_tasks` so it is directly consumable by `EQ_predict` / `EQ_evaluate`.
- Emits a sibling parquet of unique `(subject_id, prediction_time)` tuples at `{out_dir}/eval_unique/{split}/{shard}.parquet` — the input shape MEDS_EIC_AR's trajectory-generation baseline expects (one trajectory per prediction time, not one per `(code, duration)` pair).
- New `write_unique_prediction_times` config flag (default `true`) gates the sibling file.

Replaces the previously-external `dedupe_eval_tasks_for_trajectories.py` post-processing step.

## Test plan
- [x] `uv run pytest tests/test_generate_evaluation_tasks_cli.py -v` — 17 passing locally
- [ ] Cluster smoke: `EQ_generate_evaluation_tasks split=held_out input_shard=0 ...` produces both `$TASK_DIR/eval/...` and `$TASK_DIR/eval_unique/...` with expected schemas
- [ ] Confirm `EQ_predict` / `EQ_evaluate` consume the new (post-censor) labels parquet without changes

### DIFF
--- a/src/every_query/generate_tasks/configs/sample_evaluation_tasks_config.yaml
+++ b/src/every_query/generate_tasks/configs/sample_evaluation_tasks_config.yaml
@@ -53,6 +53,12 @@ durations: [30, 90, 180, 365, 731]
 # values raise ValueError.  Leave as null to disable.
 subject_subsample_fraction: null
 
+# Also emit a deduped (subject_id, prediction_time) parquet under
+# {out_dir}/eval_unique/{split}/{shard}.parquet — the input shape MEDS_EIC_AR's
+# trajectory-generation baseline expects (one trajectory per prediction time,
+# not one per (code, duration) pair). Computed after the censor filter.
+write_unique_prediction_times: true
+
 overwrite: false
 
 hydra:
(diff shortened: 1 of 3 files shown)


gkondas:
I reviewed the diff and ran the focused test plan locally on the PR branch:

`uv run pytest tests/test_generate_evaluation_tasks_cli.py -v`

Result: `17 passed`.

No blocking issues found. The implementation consistently drops censored rows before writing both the labeled eval parquet and the new `eval_unique` sibling, and the skip behavior handles the case where labels already exist but the unique sibling is missing.

One small residual test gap: the new end-to-end test checks the unique parquet column shape and row count against `labels.select(subject_id, prediction_time).unique()`, but it does not assert the exact set of rows. Given the implementation derives `unique_df` directly from that same projection, I don’t think this is blocking.

Issue #210, opened 2026-06-17
Orchestration & parallelism: resolve_workers, ProcessPoolExecutor fan-out, POLARS_MAX_THREADS pin


gkondas:
Wire the whole pipeline into one console script and fan Stage 4 out across shards.

## Scope
- `EQ_generate_training_tasks` (existing `[project.scripts]` → `sample_tasks:main`) runs Stages 0–3 inline, then the in-process Stage 4 pool — one process, one node. Multi-node out of scope.
- `resolve_workers(max_workers=None)`: cores from `SLURM_CPUS_PER_TASK` → `SLURM_CPUS_ON_NODE` → `os.cpu_count()`; `min(cores, max_workers)` if cap set (downward only).
- Stage 4 fan-out via `concurrent.futures.ProcessPoolExecutor(max_workers=resolve_workers(cfg.max_workers))`. Submit **shard ids/paths, never DataFrames** — each worker does its own parquet I/O + atomic output. `as_completed` + `fut.result()` re-raises so a failed shard aborts loudly. Driver creates `out_dir` once before the pool.
- **Pin `POLARS_MAX_THREADS=1`** at the very top of `sample_tasks.py`, **before any polars import** (workers inherit it). With 200+ shards, process fan-out already saturates cores; intra-op threads would oversubscribe (N×cores). `max_workers` is sized by RAM, not cores. Guard: any transitive `import polars` before the env line silently defeats it.
- sbatch template: single node, single task, many cpus — `--partition=cpu --nodes=1 --ntasks=1 --cpus-per-task=32`, no array/dependency/`srun` (keep `srun` for DDP training, not here). Same command runs on a research server (`os.cpu_count()` fallback).

## Rejected (do not implement)
`SLURM_NTASKS`/`srun` pool sizing (counts whole-allocation tasks, not node cores); Hydra `-m` fan-out (re-runs Stages 0–3 per shard); memory auto-clamp (peak memory is a fragile estimate — manual `--max-workers` cap instead).

## Dependencies
- Drives #STAGE4; needs #CONFIG (`max_workers`), #ARTIFACT.

Part of #202

Issue #204, opened 2026-06-17
Artifact layout: two disjoint, never-nested roots + cleanup semantics


gkondas:
Establish the two-root on-disk layout (invariant 7) that all stages read/write against.

## Scope
- `training_tasks_dir/{split}/{shard}.parquet` — **final outputs only**; at rest contains nothing but `{shard}.parquet` (directly glob-consumable, no `_`-prefixed entries).
- `training_task_artifacts_dir/{split}/` — all intermediates:
  - `_prediction_time_counts.parquet` (Stage 0)
  - `_prediction_times/{shard}.parquet` (Stage 0 map)
  - `_index/{shard}.parquet` (Stage 3)
- Roots are disjoint and never nested → cleanup is a single `rm -rf training_task_artifacts_dir` that cannot touch the dataset.
- **Atomic-write temp exception:** Stage 4 temps (`.{shard}.parquet.tmp.{pid}`) live in `training_tasks_dir/{split}/` (sibling of target) because `os.replace` requires same filesystem — not in the artifacts root. Document/enforce this placement.

## Dependencies
- Pairs with #CONFIG (the `training_task_artifacts_dir` sibling default lives there).
- Referenced by Stages 0, 3, 4.

Part of #202

Issue #203, opened 2026-06-17
Config & inputs: Hydra keys + path resolution for the redesigned sampler


gkondas:
Wire the redesign's input surface into the Hydra config and `main()` path resolution.

## Scope
- Hydra keys: `num_queries`, `num_contexts_per_query`, `min_prediction_times_per_subject` (default 50), `max_workers` (default `null`), `path_to_data`, `split`, `training_tasks_dir`, `training_task_artifacts_dir` (default `null`), `seed`.
- `min_prediction_times_per_subject` governs Stage 0 eligibility and the Stage 2 draw range (distinct prediction times, **not** events). Supersedes `min_context_per_subject`.
- `max_workers` caps `resolve_workers()` downward only (`min(cores, max_workers)`); supersedes the old `num_workers` knob.
- Path resolution: reuse the existing env-var/dotenv resolution pattern (`_resolve_path`). When `training_task_artifacts_dir` is `null`, default it to a **sibling** of `training_tasks_dir` (`{parent}/{name}_artifacts`) so the two trees never nest (invariant 7).
- `QueryDistribution.from_config(cfg, query_codes=read_query_codes(...))` plumbing — resolution stays outside the dataclass (no file I/O inside it).

## Dependencies
- Pairs with #ARTIFACT (artifact layout) for the two-root contract.
- Consumed by all stages.

Part of #202

Issue #202, opened 2026-06-17
Epic: Task sampler redesign — 5-stage sample_tasks.py pipeline


gkondas:
## Why

`sample_tasks.py` is being redesigned around a clean 5-stage pipeline that separates a single-process global driver (Stages 0–3) from an embarrassingly-parallel per-shard labeling fan-out (Stage 4). The current implementation thresholds eligibility on *events* (`min_context_per_subject`), resolves indices inside workers, and has no persisted prediction-time map. The redesign fixes the indexing space to **distinct `(subject_id, time)` rows**, resolves index→timestamp exactly once in the driver, and persists a canonical prediction-time map.

Design doc: `src/every_query/generate_tasks/redesign-spec.md`

## Pipeline

```
Stage 0  build _prediction_time_counts + _prediction_times (scan shards once, cache)
Stage 1  sample queries
Stage 2  sample (subject_idx, prediction_time_index)
Stage 3  resolve prediction_time and write per-shard index df
Stage 4  label each index shard independently
```

Stages 0–3 run in one driver process; Stage 4 fans out one worker per shard.

## Already built (do NOT re-create — build on these)

In `sample_tasks.py` / `utils/seeds.py` / `data/schema.py`:
- `read_query_codes()` — code-universe resolution (kept; stays outside `QueryDistribution`).
- `derive_seed()` (`every_query.utils.seeds`) — seed-axis splitting.
- Atomic-write helpers (`_atomic_write_parquet`, `_unique_tmp_path`, `os.replace`) and the sibling-temp pattern.
- The `join_asof` labeling core with the `+1µs` strict-after shift (`evaluate_index_df`, `compute_max_time_per_subject`) — reuse/adapt; do not reinvent the asof rule.
- `TaskQuerySchema.align()` at the write boundary.

**Being replaced:** `TaskSpec`/`sample_tasks` (→ `QueryDistribution`), event-based `sample_contexts` (→ Stage 2 distinct-time draw), `min_context_per_subject` (→ `min_prediction_times_per_subject`).

## Core invariants (from the spec)

1. One indexing space: distinct `(subject_id, time)` sorted ascending.
2. `prediction_time_index` = zero-based dense rank within a subject's distinct times.
3. Single source of truth: `(subject_id, prediction_time_index)→time` computed once (Stage 0), resolved once (Stage 3); Stage 4 does no index-space work.
4. Subjects may not span shards (hard error in Stage 0).
[...]

PR #214, opened 2026-06-17
Stage 0: build + cache prediction-time map and counts summary (#205)

### LINKED ISSUE #205: Stage 0: build + cache prediction-time map and counts summary
Scan the split's shards once and persist the canonical prediction-time indexing artifacts.

## Outputs (under `training_task_artifacts_dir/{split}/`)
- `_prediction_times/{shard}.parquet` — **canonical source of truth.** One row per eligible `(subject_id, distinct time)`: `["subject_id", "prediction_time_index", "time"]`.
- `_prediction_time_counts.parquet` — derived subject summary: `["subject_id", "shard", "n_prediction_times"]`. Exists only to make Stage 2 cheap; **not** an independent source of truth (`n_prediction_times = row count in _prediction_times/`).

## Algorithm
1. Loop shards `path_to_data/data/{split}/{i}.parquet`, read only `subject_id`,`time`; dedup to distinct `(subject_id, time)`; record each subject's `shard`.
2. Concatenate; enforce invariant 4 — `group_by(subject_id).n_unique(shard)`, any `>1` **raises** listing offenders (hard error).
3. Per subject, sort distinct times ascending → contiguous zero-based `prediction_time_index` via `pl.int_range(pl.len()).over("subject_id")` (no within-subject ties ⇒ identical to dense rank; cheaper than `.rank("dense")`). Gapless `[0,n)` is what Stage 2's array-bounded draw relies on.
4. Derive `_prediction_time_counts` from the map (per-subject `shard` + row count).
5. Eligibility `n_prediction_times >= min_prediction_times_per_subject + 1` (the `+1` keeps Stage 2's `[min, n)` range non-empty); filter both, then sort `_prediction_time_counts` by `subject_id` (row position = `subject_idx`).
- `patient_universe_size` = rows in `_prediction_time_counts`.

## Caching
Reuse both artifacts for a `(split, min_prediction_times_per_subject)` if present; `_prediction_time_counts` is valid only while it agrees with `_prediction_times/` (row-count identity).

## Dependencies
- Needs #ARTIFACT (paths) + #CONFIG.
- Feeds #STAGE2 and #STAGE3.

Part of #202

gkondas:
## Summary

Implements **Stage 0** of the task-sampler redesign (epic #202, issue #205): scan a split's shards **once** and persist the canonical prediction-time indexing artifacts the redesigned driver (Stages 1–3) and labeling fan-out (Stage 4) rely on.

New entry point `build_prediction_times(path_to_data, training_task_artifacts_dir, split, min_prediction_times_per_subject, overwrite=False) -> int`:

1. Discovers shards, reads only `subject_id`/`time`, dedups to distinct `(subject_id, time)` (`time` cast to `Datetime("us")` for the downstream `+1µs` asof rule).
2. Enforces **invariant 4** — a subject may not span shards (hard `ValueError` listing offenders).
3. Assigns a gapless zero-based `prediction_time_index` per subject via `pl.int_range(pl.len()).over("subject_id")` over time-sorted distinct rows (invariant 2).
4. Derives the subject summary, filters both to eligibility (`n >= min + 1`), sorts counts by `subject_id` (row position = `subject_idx`), and writes everything via the existing atomic-write helpers.

### Artifacts (under `training_task_artifacts_dir/{split}/`)
- `_prediction_times/{shard}.parquet` — canonical map `(subject_id, prediction_time_index, time)`.
- `_prediction_time_counts.parquet` — derived summary `(subject_id, shard, n_prediction_times)`.
- `_prediction_times_meta.json` — cache sidecar recording the `min` used (written **last** as the commit marker).

### Caching
The artifact paths don't encode `min_prediction_times_per_subject`, so a JSON sidecar records it. A cache is reused only when the sidecar's `min` matches, all files exist, and the row-count identity holds (`sum(n_prediction_times) == map rows`, `counts.height == distinct subjects`). A `min` change auto-rebuilds; `overwrite=True` forces a rebuild.

### Notes
- Does **not** rewire `main` — the driver cutover lands with #210.
- Builds on #203 (config/path resolution) and #204 (two-root artifact layout); legacy pipeline untouched.

## Testing
- `uv run pytest tests/test_sample_tasks.py -k Stage0` → 11 passed (index correctness/dedup, `+1` eligibility boundary, row-count identity, `subject_idx` ordering, shard-spanning error, per-shard partitioning, empty-split error, cache reuse/min-invalidation/overwrite).
- Full module + doctests → 77 passed, no regressions. `ruff check`/`format` clean.
- Real-data validation deferred to cluster once #210 wires Stage 0 into the driver.

Part of #202. Closes #205.

### DIFF
--- a/src/every_query/generate_tasks/redesign-spec.md
+++ b/src/every_query/generate_tasks/redesign-spec.md
@@ -46,27 +46,27 @@ flowchart TD
 These hold throughout the design; later sections reference them rather than restate them.
 
 1. **One indexing space.** Prediction times are the **distinct `(subject_id, time)` rows**, sorted
-   ascending by `(subject_id, time)` — in polars,
-   `events.select(["subject_id", "time"]).unique().sort(["subject_id", "time"])`. Every distinct event
-   time is a prediction time once the subject clears `min_prediction_times_per_subject`.
+    ascending by `(subject_id, time)` — in polars,
+    `events.select(["subject_id", "time"]).unique().sort(["subject_id", "time"])`. Every distinct event
+    time is a prediction time once the subject clears `min_prediction_times_per_subject`.
 2. **`prediction_time_index` is a zero-based dense rank** of a timestamp within its subject's sorted
-   distinct times. It equals the number of prediction times strictly before the chosen one.
-   `n_prediction_times` is the subject's count of distinct times.
+    distinct times. It equals the number of prediction times strictly before the chosen one.
+    `n_prediction_times` is the subject's count of distinct times.
 3. **Single source of truth.** The `(subject_id, prediction_time_index) → time` mapping is computed
-   **once** in Stage 0 and persisted as `_prediction_times/{shard}.parquet`. Stage 3 resolves indices to
-   timestamps by joining that map. **Stage 4 never performs index-space work** — no dedup, sort, rank,
-   or index→timestamp resolution; it receives a real `prediction_time` and labels.
+    **once** in Stage 0 and persisted as `_prediction_times/{shard}.parquet`. Stage 3 resolves indices to
+    timestamps by joining that map. **Stage 4 never performs index-space work** — no dedup, sort, rank,
+    or index→timestamp resolution; it receives a real `prediction_time` and labels.
 4. **Subjects may not span shards.** A subject lives in exactly one shard (Stage 4 derives
-   `max_time[subject_id]` from a single shard). Stage 0 enforces this as a hard error.
+    `max_time[subject_id]` from a single shard). Stage 0 enforces this as a hard error.
 5. **Determinism.** Distinct `(subject_id, time)` rows have no within-subject ties, so the dense rank is
-   unique and a given `seed`/`prediction_time_index` always maps to the same `prediction_time`. All draws
-   derive from `seed` with a fixed RNG consumption order (see *Determinism*).
+    unique and a given `seed`/`prediction_time_index` always maps to the same `prediction_time`. All draws
+    derive from `seed` with a fixed RNG consumption order (see *Determinism*).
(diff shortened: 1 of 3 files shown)


gkondas:
1) It is probably not smart to put all of the subject_id's into a python list. Should do this with a semi join instead?
```python
eligible_ids = eligible["subject_id"].to_list()
prediction_times = prediction_times.filter(
    pl.col("subject_id").is_in(eligible_ids)
)

**Comment by gkondas** [comment]
stale cache check is way too expensive and should be made less so

**Comment by gkondas** [comment]
From chat gpt: 

Nit / invariant documentation: Stage 2 relies on _prediction_time_counts.parquet having a stable row order, since subject_idx is defined implicitly as the row position of the counts table. This currently holds because Stage 0 sorts eligible by subject_id before writing, but that coupling is easy to break in a future refactor. I'd consider documenting this as an explicit invariant near the write of _prediction_time_counts.parquet (or in the design spec), e.g.

```python
# Row position defines subject_idx for Stage 2. Preserve subject_id sort order when
# writing this table; reordering rows will change the sampling universe.
```

Minor: subject_idx doesn't need to be wrapped in pl.Series; gather() can operate directly on the NumPy array returned by rng.integers(), avoiding an unnecessary temporary object.

gkondas:
Addressed part 1: added an explicit INVARIANT comment at the `_prediction_time_counts.parquet` write site noting that row position defines Stage 2's `subject_idx` and the `subject_id` sort order must be preserved on refactor.

Part 2 (the `pl.Series`/`gather()` nit): deferring — that code doesn't exist yet in this PR. `subject_idx = rng.integers(...)` and the `gather()` live only in `redesign-spec.md` as Stage 2 design. Will pass the NumPy array directly to `gather()` (no `pl.Series` wrapper) when Stage 2 is implemented.

Issue #208, opened 2026-06-17
Stage 3: resolve prediction times, zip queries, write partitioned index


gkondas:
Resolve `(subject_id, prediction_time_index) → prediction_time` once in the driver and write the Stage 4 handoff index.

## Scope
- `np.repeat` the Stage 1 queries `num_contexts_per_query` times; zip with the N Stage 2 contexts → N rows of `(subject_id, shard, prediction_time_index, code, duration_days)`.
- **Resolve `prediction_time`:** group contexts by `shard`; for each shard read its `_prediction_times/{shard}.parquet` and join on `(subject_id, prediction_time_index)`. Join **per shard** (not one global join) so the driver holds only one shard's payload-free map at a time (flat memory). Join is total (same eligibility as Stage 2's bound) — assert no nulls after as a guard.
- Partition by `shard`; write `training_task_artifacts_dir/{split}/_index/{shard}.parquet`.
- **Output columns:** `["subject_id", "prediction_time", "code", "duration_days"]` (shard implied by partition; `prediction_time_index` resolved away, not carried forward).

This is the single index-resolution point (invariant 3) — Stage 4 never resolves indices.

## Dependencies
- Needs #STAGE0 (`_prediction_times/`), #STAGE1 (queries), #STAGE2 (contexts), #ARTIFACT (paths).
- Feeds #STAGE4.

Part of #202

PR #217, opened 2026-06-18
Stage 3: resolve prediction times, zip queries, write partitioned ind…

gkondas:
…ex (#208)

Add `build_index` which zips Stage 1 queries with Stage 2 contexts, resolves `prediction_time_index → prediction_time` per shard via the Stage 0 map, and writes `_index/{shard}.parquet` — the handoff artifact consumed by Stage 4.

Claude-Session: https://claude.ai/code/session_01Q41Gr8REtDonbybdpU6f9L

### DIFF
--- a/src/every_query/generate_tasks/sample_tasks.py
+++ b/src/every_query/generate_tasks/sample_tasks.py
@@ -1330,9 +1330,7 @@ def build_index(
     # rather than materializing every shard's frame up front.
     for shard_key, shard_group in combined.group_by("shard"):
         (shard_name,) = shard_key
-        pt_map = pl.read_parquet(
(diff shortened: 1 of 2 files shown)


gkondas:
from ChatGPT:

Overall solid and easy to follow. A few suggestions:

group_by("shard") unpacking could be simplified to avoid tuple semantics.
Consider verifying/casting join key dtypes (subject_id, prediction_time_index) to prevent silent join failures.
If scale increases, repeated read_parquet per shard may become a bottleneck—worth considering caching.
Null-check after join is good; adding a small sample of offending rows would help debugging.
Optional: consider inner join if nulls are strictly invalid by design, to make the failure mode more explicit.

PR #219, opened 2026-06-19
Task sampler refactor: rewire pipeline + spawn-based Stage 4 pool (#210)

gkondas:
## Summary
- Rewires `main` for the task sampler refactor (#210)
- Cleanup of legacy code
- Stage 4 labeling pool now uses the `spawn` multiprocessing start method instead of the Linux default `fork`, fixing a polars/rayon fork deadlock in worker processes

## Notes
- Pipeline is wired up but still needs more thorough testing.


gkondas (commented):
## Review

The structural change is sound and well-executed. Collapsing the per-worker `(input_shard, task_shard)` Hydra-sweep model into a single-process driver (Stages 0–3) plus a `ProcessPoolExecutor` fan-out (Stage 4) matches the redesign spec, and the dead `sample_tasks`/`sample_contexts`/`build_index_df`/`run_worker` code + their tests are cleanly removed. The `spawn` start-method fix is correct, and the docstring explaining *why* (rayon threads holding locks across `fork`) is exactly the kind of comment that earns its keep.

A few things to address before merge.

### Must fix — leftover scratch comments

These read as notes-to-self and should be deleted (or made into real comments):

1. **End of `sample_tasks.py`** — trailing orphan after the `if __name__` guard: `# Dont let the rng go to deterministic stage as a sanity check`. References nothing; delete.
2. **`build_index` (`combined = contexts.with_columns(...)`)** — `# use hstack so its easier to read` contradicts the code (it uses `with_columns`, not `hstack`). Remove or fix.
3. **`build_prediction_times` eligibility filter** — `# Note probably can switch to > and no plus 1` above the `>= min_prediction_times_per_subject + 1` filter. This is more than cosmetic: the `+1` offset is tied to the `+1µs` strict-after asof convention, and leaving "probably can switch" invites a future contributor to make exactly the simplification that breaks the boundary semantics. Delete it, or file an issue if you genuinely want to revisit.
4. **`build_prediction_times`** — `# Notes: Write this as a test possibly?` before `_assert_no_subject_spans_shards`. Drop it or open a test issue.
5. **`sample_patient_contexts`** — `# Make this easier to read` above the `subject_id` gather is an empty placeholder.

### Worth a look — correctness

6. **Stage 4 row-count guard counts stale outputs.** In `run()`, `written = pl.scan_parquet(out_dir / "*.parquet")...` globs everything in `out_dir`, but unlike `index_dir` (rebuilt via `shutil.rmtree` in `build_index`), `out_dir` is never cleaned. On a rerun with `overwrite=False` and a different shard set / `num_queries`, leftover `{shard}.parquet` files get counted — so the guard can spuriously raise, or with `overwrite=True` and a smaller shard set, pass while including stale shards in the union. The idempotent-rerun test only covers the identical-cfg case. Consider counting only the shards this run produced, or pruning non-current shards from `out_dir`.

7. **`resolve_workers` assumes the SLURM var is a clean int.** `int(os.environ[var])` raises `ValueError` on an empty string (`SLURM_CPUS_PER_TASK=""`), which the `in os.environ` check still passes through. A guard / fall-through on non-int would be more robust.

### Nits
- `import multiprocessing` is out of alphabetical order among the stdlib imports.
- Hoisting `os` above the `POLARS_MAX_THREADS` pin is correct — keep the explanatory comment.

### Tests
`TestResolveWorkers` and `TestMainOrchestration` are a reasonable replacement and exercise the real `run()` path end-to-end including the pool fan-out. The main coverage gap is the stale-output rerun in #6.

gkondas:
Heads-up for reviewers: the red `run_tests_ubuntu` CI here is **expected** and not a blocker for merging into `dev-task-sampler` (treated as an unstable staging branch for the #202 redesign).

The 13 failing tests are CLI/e2e tests that still use the legacy `EQ_generate_training_tasks` contract (`data_dir=`, `n_tasks=`, etc.). This PR rewrote the unit tests in `test_sample_tasks.py` but intentionally leaves the CLI-driven tests for a separate follow-up PR.

Tracked in #220, which will carry both the coupled production change (hybrid override-or-env path keys, matching `sample_evaluation_tasks.py`) and the test renames, landing green on its own.

PR #231, opened 2026-06-21
Reorganize sampler tests into per-stage package, prune redundancy, add gap tests

gkondas:
## Context

The 5-stage training-task sampler (`sample_tasks.py`) already shipped with a large test suite (the redesign landed stage-by-stage across #203–#230, each with tests). This PR **does not add tests from scratch** — it audits the existing ~1.8k-line `tests/test_sample_tasks.py` monolith, reorganizes it into a per-stage package that maps 1:1 onto the pipeline, prunes a handful of low-value/redundant tests, and fills real coverage gaps.

## New layout (`tests/sampler/`)

| File | Covers | Tests |
|---|---|---|
| `test_stage0_prediction_times.py` | `build_prediction_times` (eligibility, gapless index, invariant 4, cache) | 12 |
| `test_stage1_query_distribution.py` | `QueryDistribution` | 16 |
| `test_stage2_patient_contexts.py` | `sample_patient_contexts` | 14 |
| `test_stage3_build_index.py` | `build_index` | 13 |
| `test_stage4_labeling.py` | `evaluate_index_df` + `label_one_shard` | 18 |
| `test_pure_helpers.py` | primitives, path/config resolution, atomic writes, worker sizing | 31 |
| `test_orchestration.py` | `run()` end-to-end, cross-process determinism, value snapshot | 7 |
| `conftest.py` | shared fixtures/helpers + autouse override (see below) | — |

**111 sampler tests, all passing offline.**

## Pruned (~9 tests, no coverage loss)

- Trivial path-string equalities (`test_final_output_path`, etc.) — already pinned by the source doctests.
- Redundant `resolve_training_task_paths` wrapper cases that only re-prove `_resolve_path`.
- `test_unique_tmp_path_changes_between_calls` — essentially tested `tempfile.mkstemp`.
- Brittle config-literal change-detectors (`test_legacy_keys_are_gone`, `test_defaults_match_spec`) → replaced by a presence + type check.

## Added gap tests

- **Stage 1:** log-uniform skews shorter than uniform; code draw covers the universe.
- **Stage 2:** subject draw covers the eligible universe; `prediction_time_index` dtype on the non-empty path.
- **Stage 3:** artifact-level determinism; query↔context zip alignment across shards.
- **Stage 4:** event *before* prediction_time isn't an occurrence (forward-only asof); non-matching code in window ignored (query isolation); multiple matches still `True`.
- **Orchestration:** the row-count guard actually *raises* on a budget mismatch (previously only the happy path was tested); `overwrite=True` forces a rebuild.
- Strengthened the atomic-write test to assert the failure path leaves no orphan temp.

## Decoupling from the HF model fixture

The sampler stages are pure data transforms, but the repo-root `autouse` `_setup_doctest_namespace` fixture pulls in `demo_model` (which builds `answerdotai/ModernBERT-base` via `AutoConfig.from_pretrained`), forcing every test — including pure sampler units — to require HuggingFace/network access. `tests/sampler/conftest.py` overrides that autouse fixture so the sampler layer runs fast and offline. The one test that genuinely needs the model (downstream dataset drop-in) moves to top-level `tests/test_sampler_dataset_integration.py`, where the root fixtures still apply.

## Validation

- ✅ `pytest tests/sampler` — **111 passed** (offline).
- ⏳ Not run in the dev sandbox (need HF access — please confirm green in CI): `tests/test_generate_tasks.py`, `tests/test_e2e_foundation.py`, `tests/test_sampler_dataset_integration.py`, and source `--doctest-modules`. These are unchanged by this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SXUwK16bdaei5fMsC4bMYi)_


gkondas (commented):
## Review

Nice reorganization — the per-stage split maps cleanly onto the pipeline, the gap tests are well-chosen (forward-only asof, query-code isolation, log-uniform shape, the row-count guard actually *raising*, atomic-write orphan cleanup), and the `conftest` override that decouples the pure stages from the HF model fixture is well-reasoned. I ran the suite locally: **`pytest tests/sampler` → 111 passed**, matching the description. I also spot-checked the pruning claim — the ~10 dropped tests are genuinely redundant (the path-string equalities are pinned by the source doctests, which CI runs via `--doctest-modules`), so I agree there's no real coverage loss.

That said, **CI is red on all three checks**, and two are caused by this PR (not the HF-access gap the description anticipated). Both reproduce locally and have small fixes:

### 🔴 Blocking 1 — test collection errors (`run_tests_ubuntu` 3.11 & 3.12)

11 collection errors, all from:
```
ImportError: cannot import name 'ENSURE_ENV_PLACEHOLDERS' from 'conftest'
(/home/runner/.../tests/sampler/conftest.py)
```
Root cause: `tests/sampler/` has **no `__init__.py`**, unlike `tests/` and `tests/training_validity/` which both do. In pytest's default `prepend` import mode, the new `tests/sampler/conftest.py` therefore gets imported under the bare top-level name `conftest` and claims `sys.modules['conftest']`. `tests/training_validity/test_training_validity.py:20` then does `from conftest import ENSURE_ENV_PLACEHOLDERS, run_and_check` and resolves to the *sampler* conftest, which doesn't define those names → the whole suite (`test_generate_tasks`, `test_process_data`, `test_train`, …) errors out at collection. This is why the "confirm green in CI" tests didn't actually exercise HF — they never got past collection.

**Fix:** add an empty `tests/sampler/__init__.py` (consistent with the rest of the tests tree). I verified locally that this makes the sampler conftest import as `tests.sampler.conftest` and collection then succeeds:
```
$ touch tests/sampler/__init__.py
$ pytest tests/sampler tests/training_validity --collect-only -q
112/113 tests collected
```

### 🔴 Blocking 2 — `code-quality`

`ruff format --check` fails (locally on `tests/sampler/test_stage0_prediction_times.py`; CI's ruff also flags stage2/stage3, likely a version skew). Run `ruff format tests/sampler/` and commit.

Once those two are addressed the PR looks good to me. The snapshot test pinning literal float32 values (`EXPECTED_SNAPSHOT_ROWS`) is a deliberate regression guard but will need a refresh on any numpy-Generator/polars change — worth a one-line note that it's expected to be regenerated, which you already include in the assertion message. 👍

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #234, opened 2026-06-22
Fix #184: post-compose config validation, retire ensure_env() presence gate

### LINKED ISSUE #184: ensure_env() gating: drop dead vars, validate resolved Hydra config instead of blind env-var presence
## Problem

`every_query.utils._env.ensure_env()` is a blind env-var **presence** gate. `train.py::_init_env()` calls it unconditionally before the Hydra config is touched:

```python
REQUIRED_ENV_VARS = ("PROJECT_DIR", "OUTPUT_DIR", "TASK_DIR", "FINAL_DATA_DIR", "WANDB_ENTITY")
```

Three problems, all confirmed against `origin/dev` (`_env.py` is byte-identical to `main`; `train.py::_init_env` still calls `ensure_env()` unconditionally).

### 1. `PROJECT_DIR` is dead

No Hydra config interpolates `${oc.env:PROJECT_DIR}`; no Python reads it. `.env.example` uses it only to derive `SAVE_DIR` / `OUTPUT_DIR` *within that file*. `ensure_env()` requires it non-empty purely as theater.

Adjacent dead/wrong vars in `.env.example` (not gated, but worth cleaning in the same pass): `SAVE_DIR` has zero usages anywhere; `INTERMEDIATE_DIR` is explicitly labelled "duplicate kept for backward compatibility" with zero usages; `RAW` is a naming mismatch — preprocessing reads `RAW_MEDS_DIR`, which `preprocessing/__main__.py` sets internally from the `input_dir=` CLI arg, so `RAW` in `.env.example` is never read.

### 2. The 4 "real" vars are only needed if you *don't* override the corresponding Hydra node

Each backs exactly one config node:

| Env var | Config node |
|---|---|
| `FINAL_DATA_DIR` | `datamodule.config.tensorized_cohort_dir` |
| `TASK_DIR` | `datamodule.config.task_labels_dir` |
| `OUTPUT_DIR` | `output_dir` (`${oc.env:OUTPUT_DIR}/${run_id:}/`) |
| `WANDB_ENTITY` | `trainer.logger.entity` |

A CLI override — `datamodule.config.tensorized_cohort_dir=/path` — replaces the node entirely. OmegaConf interpolations are lazy, so `${oc.env:FINAL_DATA_DIR}` is then never evaluated. But `ensure_env()` runs first and has no idea you overrode anything, so it still hard-requires the env var to be non-empty. Result: you export junk values just to clear the gate, and the gate gives a false impression that those values must be valid.

A fully CLI-driven invocation today still needs:

```bash
FINAL_DATA_DIR=x TASK_DIR=x OUTPUT_DIR=x WANDB_ENTITY=x PROJECT_DIR=x \
EQ_train datamodule.config.tensorized_cohort_dir=/real/... output_dir=/real/... ...
```

— every env var there is decoration.

### 3. `WANDB_ENTITY` is required even when the logger is disabled

`trainer.logger=false` (what `_demo_train.yaml` does) means `${oc.env:WANDB_ENTITY}` is never evaluated — but `ensure_env()` still demands it.

## Why the gate exists

Per the `_env.py` docstring: it turns a cryptic mid-run `InterpolationResolutionError` (`${oc.env:X}` with `X` unset) into one clear early error. The intent is good; the implementation is too blunt — it validates env-var *presence* rather than *resolved-config validity*.

(Side note: the docstring claims it's "imported early by both `train.py` and `tasks.py`" — `tasks.py` no longer exists / no longer calls it; `train.py::_init_env` is the only caller. Docstring is stale.)

## Proposed fix

Move validation from **"env-var presence, pre-compose"** to **"resolved-config validity, post-compose"**:

1. **Drop `PROJECT_DIR`** from `REQUIRED_ENV_VARS` (dead). Clean `SAVE_DIR` + `INTERMEDIATE_DIR` from `.env.example`, and fix the `RAW` → `RAW_MEDS_DIR` naming note.
2. **Null-default the env interpolations** in `config.yaml`: `${oc.env:FINAL_DATA_DIR,null}`, `${oc.env:TASK_DIR,null}`, `${oc.env:WANDB_ENTITY,null}`. A missing env var + missing override then resolves to `null` instead of crashing — and is caught by step 3 with a clear message.
3. **Add a post-compose `validate_training_config(cfg)`** in `train.py`, called from `main()` once the config is available:
   - `tensorized_cohort_dir` / `task_labels_dir` — must resolve to existing directories (read inputs).
   - `output_dir` — must resolve to a non-empty path (write target; training creates it, so don't require pre-existence).
   - wandb `entity` — validated **only if** `trainer.logger` is wandb-shaped; skipped when `logger` is `false` / `null` / a non-wandb logger.
4. **`ensure_env()`** keeps only `load_dotenv()` (still needed so `.env` values are visible to the lazy `${oc.env:...}` interpolations) — the blind presence check is removed.

Net effect: a fully CLI-driven `EQ_train` needs **zero** env vars; an `.env`-driven run still gets a clear early error if a path is missing or wrong; `trainer.logger=false` no longer needs a phantom `WANDB_ENTITY`.

## Related

- #117 — the original `REQUIRED_ENV_VARS` audit that removed `PROCESSED` / `INTERMEDIATE` from the required list. This is the natural follow-up: that issue trimmed the list; this one fixes the *mechanism*.

gkondas:
Closes #184.

## Problem
`ensure_env()` was a blind env-var **presence** gate run *before* the Hydra config was composed. It couldn't see CLI overrides (a `datamodule.config.task_labels_dir=/path` override makes the backing `${oc.env:TASK_DIR}` interpolation never evaluate, yet the gate still demanded the env var), required `WANDB_ENTITY` even when the logger was disabled, and hard-required the dead `PROJECT_DIR`. Separately, the redesigned sampler writes to `$TRAINING_TASKS_DIR` but train still defaulted `task_labels_dir` to `$TASK_DIR`, so `EQ_train` did not auto-read sampler output.

## Changes
- **`utils/_env.py`** — `ensure_env()` now only `load_dotenv()`; `REQUIRED_ENV_VARS` + presence loop removed; stale docstring rewritten.
- **`train/train.py`** — new `validate_training_config(cfg)` (+ `_is_wandb_logger` helper), called from `main()` after `_init_env()`. Validates the **resolved** config: cohort/task dirs must exist, `output_dir` non-empty, wandb `entity` required only for a wandb-shaped logger.
- **`train/configs/config.yaml`** — null-default the `${oc.env:...}` interpolations; **rewire** `task_labels_dir` from `TASK_DIR` → `TRAINING_TASKS_DIR` so `EQ_train` auto-reads sampler output.
- **`.env.example`** — drop dead `PROJECT_DIR`/`SAVE_DIR`/`INTERMEDIATE_DIR`/`RAW`; clarify `TASK_DIR` is eval-only. (`INTERMEDIATE` stays — it's the live `data_dir` fallback for both samplers.)
- **READMEs** (top + `train/`) — update env-var docs and the `task_labels_dir` default.
- **tests** — drop `ENSURE_ENV_PLACEHOLDERS` scaffolding from conftest + 4 files; `test_train_cli.py` now runs with **zero** EveryQuery env vars (regression guard); add `tests/test_env_validation.py` (8 cases).

## Net effect
A fully CLI-driven `EQ_train` needs **zero** env vars; an `.env`-driven run gets a clear early error if a path is missing/wrong (not a cryptic `InterpolationResolutionError`); `trainer.logger=false` needs no phantom `WANDB_ENTITY`; a default `.env` run reads the sampler output automatically.

## Verification
- `test_env_validation.py` + `train.py` doctests → 12 passed
- `test_cli_smoke.py` + `test_train_cli.py` → 17 passed (zero-env subprocess train)
- `ruff` clean; manual negative check raises a clear `ValueError` naming the node + env var

> Base is `dev-task-sampler` (the sampler-redesign integration branch), not `main`.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -9,20 +9,18 @@
 
 # --- Paths -------------------------------------------------------------------
 
-# Absolute path to this repo on the current machine.
-PROJECT_DIR="/path/to/EveryQuery"
-
 # Root of the MEDS cohort. The INTERMEDIATE / PROCESSED dirs below are written
-# here by the preprocessing pipeline.
+# here by the preprocessing pipeline.  (Preprocessing reads the *raw* MEDS via the
+# `input_dir=` CLI arg and sets `RAW_MEDS_DIR` internally — there is no `RAW` env var.)
 DATA_DIR="/path/to/MIMIC_MEDS/MEDS_cohort"
 
-RAW="${DATA_DIR}"
 INTERMEDIATE="${DATA_DIR}/intermediate"
-INTERMEDIATE_DIR="${INTERMEDIATE}"   # duplicate; kept for backward compatibility
 PROCESSED="${DATA_DIR}/processed"
 FINAL_DATA_DIR="${PROCESSED}"
 
-# Where task parquet files live / are written.
+# Where *evaluation* task parquet files are written.  Only the evaluation generator
+# (`EQ_generate_evaluation_tasks`, output under `$TASK_DIR/eval/...`) reads this.
+# Training reads `TRAINING_TASKS_DIR` below, not this.
 TASK_DIR="/path/to/eq_stuff/tasks"
 
 # Final-output root for the redesigned training-task sampler (EQ_generate_training_tasks).
@@ -35,9 +33,8 @@ TRAINING_TASKS_DIR="/path/to/eq_stuff/training_tasks"
 # ACES_SHARDS_DIR="/path/to/eic_stuff/make_index_dfs/task_configs"
 
 # Where training outputs and checkpoints land. Hydra creates
-# ${OUTPUT_DIR}/outputs/<date>/<time>/ run dirs underneath.
-SAVE_DIR="${PROJECT_DIR}/results/"
-OUTPUT_DIR="${PROJECT_DIR}/results"
+# ${OUTPUT_DIR}/<date>/<time>/ run dirs underneath.
+OUTPUT_DIR="/path/to/EveryQuery/results"
 
 # --- Weights & Biases --------------------------------------------------------
 # `wandb` itself also picks these up automatically.
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ Both task-generation endpoints emit `TaskQuerySchema`-conformant parquets. Train
 
 ```bash
 EQ_process_data \
-	input_dir="$RAW" \
+	input_dir="$DATA_DIR" \
 	intermediate_dir="$INTERMEDIATE" \
 	output_dir="$FINAL_DATA_DIR"
 ```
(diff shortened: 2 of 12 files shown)


gkondas (commented):
## Review

Clean, well-scoped fix. Replacing a blind pre-compose presence gate with a post-compose check against the *resolved* config is the right model — it correctly stops demanding env vars that a CLI override makes inert, and `WANDB_ENTITY` is now scoped to an actually-enabled wandb logger. Test coverage is good (8 unit cases + doctests + the zero-env subprocess regression guard), and the `.env.example`/README cleanup matches the code. I ran `test_env_validation.py` + `train.py` doctests locally: **12 passed**.

### One real gap: the `output_dir` check never fires for the shipped config

`validate_training_config` guards `output_dir` with `if not cfg.get("output_dir")`, but in `config.yaml` the value is:

```yaml
output_dir: ${oc.env:OUTPUT_DIR,null}/${run_id:}/
```

With `OUTPUT_DIR` unset, the `null` default is concatenated into the surrounding string, so it resolves to a **truthy** `'None/<run_id>/'` rather than to `None`. Confirmed locally:

```
resolved output_dir = 'None/RUN/'
truthy? True
```

So a genuinely missing `OUTPUT_DIR` slips past the gate and you get a run directory literally named `None/...` — exactly the cryptic-failure class this PR set out to eliminate. The `tensorized_cohort_dir` / `task_labels_dir` checks don't have this problem because they're bare `${oc.env:...,null}` interpolations that resolve cleanly to `None`.

Suggestion: validate the `OUTPUT_DIR` presence directly, or reject an `output_dir` that begins with `None/` / `null/`. (Low severity — it's a write target created later, not a read input — but it undercuts the stated "clear early error" guarantee, so worth closing.)

### Minor

- `_is_wandb_logger` keys off the substring `"WandbLogger"` in `_target_`. Fine for the shipped `pytorch_lightning.loggers.wandb.WandbLogger`, but a custom subclass not named `*WandbLogger` would skip the entity check. Acceptable given current usage — just noting the assumption.

Otherwise LGTM once the `output_dir` gap is addressed (or consciously waived).

PR #239, opened 2026-06-22
Fix four bugs in the 5-stage task sampler

gkondas:
## Summary

Four bugs found while reviewing `sample_tasks.py` (the 5-stage sampling-first task generator), all fixed with tests.

### 1. Temp cleanup glob never matched real temp files
`_unique_tmp_path` built temps as `.{name}.<random>.tmp` (via `mkstemp(suffix=".tmp")`), but `_clean_stale_temps` globbed `.{name}.tmp.*` — the two never agreed, so orphaned atomic-write temps from crashed workers were **never cleaned**. The existing tests baked in the same wrong pattern, so they passed against the bug. Fix: move `tmp` into the prefix (`.{name}.tmp.<random>`), aligning real temp names with the cleanup glob, the docstrings, and the tests. Still hidden and not ending in `.parquet`, so the final-root `{split}/*.parquet` safety glob still skips it (invariant 7).

### 2. `overwrite=False` kept stale labels after a config change
Stage 3 always rebuilds the index, but Stage 4 skipped purely on output existence. Rerunning with a changed `seed`/`num_queries`/`query_codes`/duration params (same shard set) silently kept the **previous run's labels**, and the row-count guard could still pass. Fix: add a serialization-independent per-shard index fingerprint (`_index_fingerprint`, built from `hash_rows`) recorded in a `_labeled/{shard}.json` sidecar in the **artifacts** root (preserves invariant 7). Stage 4 skips only on a fingerprint match, else relabels; missing/corrupt fingerprint → relabel (safe default). Preserves rerun idempotency and `overwrite=True` semantics.

### 3. `num_queries=0` crashed the final/summary scans
The empty-budget path wrote no output, so `pl.scan_parquet(out_dir/"*.parquet")` raised "no files found" instead of reporting 0 rows. Fix: resolve an explicit file list; report 0 rows (passing the `0 == 0` guard) and return before the summary scan.

### 4. Misleading `group_by` ordering comment in `build_index`
A comment claimed `sort("shard")` made shard processing order deterministic, but `group_by` is not order-preserving without `maintain_order=True`. Output content was unaffected; added `maintain_order=True` so the comment is true.

## Tests
- `test_pure_helpers.py`: regression pinning `_unique_tmp_path` naming ↔ cleanup glob.
- `test_stage4_labeling.py`: real-temp cleanup; fingerprint skip / relabel-on-change / relabel-on-missing.
- `test_orchestration.py`: changed-seed-relabels end-to-end; `num_queries=0` no-crash.

All pass: 120 sampler unit tests, 10 module doctests, and the dataset-integration test.

### DIFF
--- a/src/every_query/generate_tasks/sample_tasks.py
+++ b/src/every_query/generate_tasks/sample_tasks.py
@@ -547,9 +547,14 @@ def _unique_tmp_path(fp: Path) -> Path:
     at the end to be the atomicity primitive.
     """
     fp.parent.mkdir(parents=True, exist_ok=True)
+    # The random token sits *after* ``.tmp.`` (prefix=".{name}.tmp.", empty suffix) so the produced
+    # name is ``.{name}.tmp.<random>`` — matching the ``.{shard}.parquet.tmp.*`` glob in
+    # ``_clean_stale_temps``.  Putting ``tmp`` in a ``suffix=".tmp"`` instead would emit
+    # ``.{name}.<random>.tmp``, which that glob never matches, so orphaned temps would never be
+    # cleaned (the random token lands between ``.parquet.`` and ``.tmp``).
     fd, tmp_name = tempfile.mkstemp(
-        prefix=f".{fp.name}.",
-        suffix=".tmp",
+        prefix=f".{fp.name}.tmp.",
+        suffix="",
         dir=str(fp.parent),
     )
     os.close(fd)
@@ -695,6 +700,7 @@ PREDICTION_TIME_COUNTS_NAME = "_prediction_time_counts.parquet"
 PREDICTION_TIMES_DIRNAME = "_prediction_times"
 PREDICTION_TIMES_META_NAME = "_prediction_times_meta.json"
 INDEX_DIRNAME = "_index"
+LABELED_DIRNAME = "_labeled"
 
 
 def final_output_path(training_tasks_dir: Path, split: str, shard: str) -> Path:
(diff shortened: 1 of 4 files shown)


gkondas:
## Multi-agent review summary

Ran three independent review agents (correctness, fingerprint-guard design, test coverage). **Verdict: no blocking bugs — the four fixes are sound and fail-safe.** Every failure mode in the new fingerprint guard biases toward over-relabeling, never wrong-skipping (unknown/corrupt/torn sidecar → relabel; collision → bounded over-skip only on content that already hashed *and* counted identically). Invariant 7, spawn-pool concurrency, and the `num_queries=0`/`num_contexts=0` guard all check out. The `_labeled/{shard}.json` sidecar is consistent with Stage 0's existing `_prediction_times_meta.json` pattern rather than a new idiom.

### Worth acting on (none blocking)

**Test gaps (highest value — both reachable with the existing `_two_shard_cohort` fixture):**
1. **`run()` sidecar pruning is untested** — the new loop deleting stale `_labeled/*.json` for dropped shards has zero coverage. A shard-count-shrinking rerun should leave no orphan sidecar.
2. **Multi-shard / fan-out changed-config relabel** — the only end-to-end fingerprint test is single-shard with `max_workers=1`, so the skip/relabel check inside *spawned* workers is never exercised.
3. **Direct `_index_fingerprint` unit test** — nothing pins its three load-bearing claims (order-independence, value-sensitivity, row-count-sensitivity).

**Code/doc nits (minor):**
4. `label_one_shard` reconstructs the sidecar path inline instead of calling the new `labeled_fingerprint_path` helper — drift risk.
5. `hash_rows()` is a polars-version-tied cache key (an upgrade silently relabels everything once — safe direction). Worth a docstring sentence that the fingerprint is **event-data-blind** like Stage 0's cache key, so a MEDS data refresh under a fixed index needs `overwrite=True`.
6. The `& 0xFFFFFFFFFFFFFFFF` mask is a no-op on an already-UInt64 sum — harmless, can drop.

**Residual collision note:** `height:hash_sum` is a weak combiner — two different multisets with equal count and equal 64-bit hash-sum would collide. Astronomically unlikely for real config changes (which alter `seed`/`num_queries` and perturb hashes broadly) and the consequence is bounded, but it's "unlikely," not "impossible."

PR #243, opened 2026-06-23
Fix #242: inline max_time computation into evaluate_index_df

### LINKED ISSUE #242: Inline max_time computation into evaluate_index_df (drop the separate arg)
## Problem

The latest-event-per-subject frame (`max_time`) used for the censoring check is computed by
`compute_max_time_per_subject(events_df)` and then **passed into** the labeling function
`evaluate_index_df`. The docstring frames this compute-then-pass split as computing the frame
"once per shard to turn the per-row censoring check into an O(1) lookup"
(`src/every_query/generate_tasks/sample_tasks.py#L345-L346`).

In practice that reuse benefit is never realized — **every** call site computes the frame once
and feeds it straight into a single `evaluate_index_df` call (1:1 ratio, no reuse across the
shard):

- `compute_max_time_per_subject` helper: `sample_tasks.py#L342-L350`
- `evaluate_index_df` signature: `sample_tasks.py#L353-L357`
- Call site (Stage 4 worker `label_one_shard`): `sample_tasks.py#L874-L877`
- Call site (evaluation tasks): `sample_evaluation_tasks.py#L377-L378`

So the split buys nothing at runtime and adds boilerplate plus an easy-to-misuse two-step
contract at each caller. `max_time` is always a pure function of `events_df`; there is no
production path where the two diverge.

## Proposal

Compute `max_time` **inside** `evaluate_index_df` and drop it from the signature:

```python
def evaluate_index_df(
    index_df: pl.DataFrame,
    events_df: pl.DataFrame,
) -> pl.DataFrame:
    ...
    max_time_per_subject = compute_max_time_per_subject(events_df)
```

Then drop the explicit pre-compute at the production call sites
(`sample_tasks.py#L874-L877`, `sample_evaluation_tasks.py#L377-L378`) →
`evaluate_index_df(index_df, events_df)`. The empty-input fast-path guard in
`sample_evaluation_tasks.py` stays; only the `max_time_df` line collapses.

Keep `compute_max_time_per_subject` as a module-level helper (now an implementation detail of
`evaluate_index_df`; it still has a direct unit test in `tests/sampler/test_pure_helpers.py`).

## Why full inlining (not an optional `max_time` arg)

An earlier idea was to make `max_time_per_subject` an *optional* argument (compute internally when
`None`) to preserve a "test injection" seam. On inspection that seam is **unused and undesirable**:

- No test passes a `max_time` frame that diverges from its `events` — every test does
  `max_time_df = compute_max_time_per_subject(events)` immediately before the call
  (`tests/sampler/test_stage4_labeling.py#L71-L78`, `#L146-L148`;
  `tests/test_sampler_dataset_integration.py#L90-L91`).
- Tests control censoring the *right* way — by steering the `events` the frame is derived from
  (e.g. adding a far-future sentinel event, `test_stage4_labeling.py#L101`), keeping `events` and
  `max_time` consistent.
- A divergent injected `max_time` would assert behavior on a state that can never occur in
  production — an anti-pattern, not a feature worth preserving.

So there is nothing to keep. Full inlining leaves the simplest two-argument function with no
vestigial seam.

## Scope / acceptance

- [ ] `max_time` computed internally in `evaluate_index_df`; parameter removed.
- [ ] Production call sites simplified to the two-arg form.
- [ ] Tests updated to drop the now-redundant `max_time_df` line and call
      `evaluate_index_df(index_df, events)` (`tests/sampler/test_stage4_labeling.py`,
      `tests/test_sampler_dataset_integration.py`).
- [ ] `compute_max_time_per_subject` kept as a helper with its existing unit test
      (`tests/sampler/test_pure_helpers.py`).
- [ ] Docstring updated to drop the stale "O(1) lookup once per shard" framing.
- [ ] `uv run pytest tests/sampler/ tests/test_generate_tasks.py tests/test_sampler_dataset_integration.py`
      passes.

## Out of scope

No change to label semantics, censoring logic, or schemas.

gkondas:
## Summary

Inlines the per-subject `max_time` computation into `evaluate_index_df`, dropping it from the signature (closes #242).

The `compute_max_time_per_subject(events_df)` → pass-into-`evaluate_index_df` split never realized its claimed "compute once per shard for O(1) lookup" reuse benefit — every call site was a 1:1 compute-then-pass. `max_time` is always a pure function of `events_df` with no production path where the two diverge, so the split only added boilerplate and a two-step contract that was easy to misuse.

## Changes

- `evaluate_index_df(index_df, events_df)` now computes `max_time` internally (after the empty-input fast-path guard).
- Production call sites simplified to the two-arg form: `label_one_shard` (`sample_tasks.py`) and `run_eval` (`sample_evaluation_tasks.py`); dropped the now-unused `compute_max_time_per_subject` import in the latter.
- `compute_max_time_per_subject` kept as a module-level helper (still directly unit-tested in `test_pure_helpers.py`); stale "O(1) lookup once per shard" docstring framing updated.
- Tests drop the redundant `max_time_df` argument (`test_stage4_labeling.py`, `test_sampler_dataset_integration.py`).

**No change to label semantics, censoring logic, or schemas.**

## Test plan

`uv run pytest tests/sampler/ tests/test_generate_tasks.py tests/test_sampler_dataset_integration.py` — **124 passed**.

### DIFF
--- a/src/every_query/generate_tasks/sample_evaluation_tasks.py
+++ b/src/every_query/generate_tasks/sample_evaluation_tasks.py
@@ -38,7 +38,6 @@ from every_query.generate_tasks.sample_tasks import (
     _atomic_write_parquet,
     _read_event_shard,
     _resolve_path,
-    compute_max_time_per_subject,
     evaluate_index_df,
     read_query_codes,
 )
@@ -374,8 +373,7 @@ def run_worker(
         ]
         labeled = empty_task_query_df().select(out_cols)
     else:
-        max_time_df = compute_max_time_per_subject(events_df)
-        labeled = evaluate_index_df(index_df, events_df, max_time_df)
+        labeled = evaluate_index_df(index_df, events_df)
 
     # Censored rows (null boolean_value) are not actionable downstream — drop them
     # so EQ_predict / EQ_evaluate consume a ready-to-score parquet.
--- a/src/every_query/generate_tasks/sample_tasks.py
+++ b/src/every_query/generate_tasks/sample_tasks.py
@@ -342,8 +342,8 @@ def sample_patient_contexts(
 def compute_max_time_per_subject(events_df: pl.DataFrame) -> pl.DataFrame:
     """Return a ``(subject_id, max_time)`` DataFrame.
 
-    Used once per shard to turn the per-row censoring check into an O(1) lookup inside
-    ``evaluate_index_df``.
+    Implementation detail of ``evaluate_index_df``'s censoring check (kept module-level for
+    direct unit testing).
     """
     return events_df.group_by(DataSchema.subject_id_name).agg(
         pl.col(DataSchema.time_name).max().alias("max_time")
@@ -353,7 +353,6 @@ def compute_max_time_per_subject(events_df: pl.DataFrame) -> pl.DataFrame:
 def evaluate_index_df(
     index_df: pl.DataFrame,
     events_df: pl.DataFrame,
-    max_time_per_subject: pl.DataFrame,
 ) -> pl.DataFrame:
     """Label an index DataFrame with the single nullable ``boolean_value`` column via a single ``join_asof``.
 
@@ -375,7 +374,6 @@ def evaluate_index_df(
             ``query``, ``duration_days``. If ``task_id`` is present it is ignored and dropped from
             the output.
         events_df: Shard events with columns ``subject_id``, ``time``, ``code``.
-        max_time_per_subject: Output of ``compute_max_time_per_subject``.
 
     Returns:
         DataFrame with columns ``(subject_id, prediction_time, boolean_value, query,
@@ -398,6 +396,8 @@ def evaluate_index_df(
     # dtypes, guaranteed to pass ``TaskQuerySchema.align()`` downstream.
     if index_df.height == 0:
         return empty_task_query_df().select(out_cols)
+
+    max_time_per_subject = compute_max_time_per_subject(events_df)
 
     # Left side: index rows with a +1µs-shifted prediction_time for the strict-> asof key.
     left = index_df.with_columns(
@@ -872,9 +872,8 @@ def label_one_shard(
     _clean_stale_temps(out_dir, shard)
 
     events_df = _read_event_shard(data_dir / f"{shard}.parquet")
-    max_time = compute_max_time_per_subject(events_df)
 
-    labeled = evaluate_index_df(index_df, events_df, max_time)
+    labeled = evaluate_index_df(index_df, events_df)
     aligned = TaskQuerySchema.align(labeled.to_arrow())
 
     _atomic_write_parquet(pl.from_arrow(aligned), final)
--- a/tests/sampler/test_stage4_labeling.py
+++ b/tests/sampler/test_stage4_labeling.py
@@ -14,7 +14,6 @@ import polars as pl
 from every_query.generate_tasks import sample_tasks as st
 from every_query.generate_tasks.sample_tasks import (
     _clean_stale_temps,
-    compute_max_time_per_subject,
     evaluate_index_df,
     label_one_shard,
 )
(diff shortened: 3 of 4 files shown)


gkondas:
**Correctness review (automated):** Verified behavior-preserving. Confirmed:

- The inlined compute at `sample_tasks.py:400` sits *after* the empty-input guard (`:397-398`) and *before* the join that consumes it (`:423`), which is unconditionally reached whenever line 400 runs — no scope/ordering issue.
- Empty `index_df` correctly returns at `:398` *before* computing `max_time_per_subject`. This is actually a tiny improvement: the old `label_one_shard` computed `compute_max_time_per_subject(events_df)` unconditionally even for empty index_df; `max_time` was always unused in that case, so output is identical and the wasted group_by is now skipped.
- `evaluate_index_df` has no other call sites in `src/` passing 3 args — grep shows only the two production sites (`sample_tasks.py:876`, `sample_evaluation_tasks.py:376`), both now 2-arg, plus docstring/comment references.
- Eval-tasks empty guard (`sample_evaluation_tasks.py:366`) is unchanged; the old code already computed `max_time_df` only inside the non-empty `else` branch, so collapsing that line is exact.
- `max_time = f(events_df)` purely, so internalizing it changes nothing semantically; censoring/occurrence logic (`:439-447`) untouched.

LGTM — no correctness concerns.

gkondas:
**Quality review (automated):** Looks clean — meets the issue's acceptance criteria with no quality concerns.

- Stale framing removed: `sample_tasks.py:345-346` reworords the "O(1) lookup once per shard" line to "Implementation detail of `evaluate_index_df`'s censoring check (kept module-level for direct unit testing)" — exactly the intent of #242.
- `Args:` block (`sample_tasks.py:372-376`) drops the removed `max_time_per_subject` param and still documents `index_df`/`events_df` accurately.
- No dangling 3-arg references: grepped `max_time_per_subject`/`evaluate_index_df` across `src/`. All remaining hits are valid (call sites now 2-arg, plus internal use of the local `max_time_per_subject` var at `sample_tasks.py:423-425`, and prose references in README/comments). No stale signatures.
- `compute_max_time_per_subject` stays module-level/public and its unit test is retained (`tests/sampler/test_pure_helpers.py:93`).
- Inserted compute (`sample_tasks.py:400`) sits correctly after the empty-input early-return (`:397-398`), so the empty path still short-circuits without touching `events_df`; reads cleanly.
- Scope discipline holds: only the param removal + call-site collapse; semantics, censoring, and schemas untouched.

gkondas:
**Test review (automated):** Coverage and test quality look good — approving from a test perspective.

**Import hygiene (verified):**
- `compute_max_time_per_subject` import correctly dropped where it became unused: `tests/sampler/test_stage4_labeling.py:17` and `tests/test_sampler_dataset_integration.py:19`. No orphaned `max_time_df = ...` lines remain in either file (all 9 call sites in test_stage4_labeling now use the two-arg form).
- Kept where still used: `tests/sampler/test_pure_helpers.py:25` retains the import and a direct unit test `test_compute_max_time_per_subject` (`:93`) for the still-public helper. Good — the helper stays exercised in its own right.

**Oracle tests correctly untouched:** `tests/test_generate_tasks.py` (`:117-118`) and `tests/training_validity/test_training_validity.py` (`:151-152`) build their own `max_time` via `events.group_by(...)` and never call `evaluate_index_df` with the dropped arg, so the refactor doesn't touch them. Correct.

**Newly-internalized compute path is exercised both ways:**
- Non-empty: all functional cases in `test_stage4_labeling.py` flow through `evaluate_index_df`'s internal `compute_max_time_per_subject` (`sample_tasks.py:400`).
- Empty index_df: `test_empty_index_partition` (`test_stage4_labeling.py:635`) drives the `height == 0` early-return (which sits *before* the internal compute, so the empty path correctly skips it) end-to-end via `label_one_shard`.

**Censoring still covered:** unknown-subject censoring (`test_unknown_subject_is_treated_as_censored`, `:123`), `test_no_occurrence_and_censored_is_null` (`:309`), occurrence-over-censoring priority (`:290`), and the end-to-end three-valued `test_censoring_logic` (`:578`). Tests steer censoring through the `events` frame (far-future sentinels), consistent with the issue's rationale for dropping the inject seam.

Ran `tests/sampler/test_stage4_labeling.py tests/sampler/test_pure_helpers.py tests/test_sampler_dataset_integration.py` — 57 passed. No substantive gaps found.

PR #245, opened 2026-06-23
Fix #235: MEICAR-style path handling (Hydra args, drop .env/load_dotenv)

### LINKED ISSUE #235: Shift path/config handling to MEDS_EIC_AR style (Hydra args + oc.env) instead of .env + load_dotenv
## Summary

Shift EveryQuery's path/configuration handling away from the `.env` + `load_dotenv()` + env-var-fallback pattern toward the upstream [MEDS_EIC_AR](https://github.com/mmcdermott/MEDS_EIC_AR) style, where path roots are passed purely as Hydra args and env vars are an internal subprocess-plumbing detail rather than a user-facing config surface.

This is a **direction-setting / eventual** change, not an urgent fix — filing it so the design is captured and we can sequence it against the sampler redesign work.

## Background: how the two projects differ

**EveryQuery (current):** path roots are real, package-visible env vars read inside the package.

- `sample_tasks.py` `main()` calls `load_dotenv()`.
- `_resolve_path()` resolves each root with **override > env var > raise** (e.g. `cfg.data_dir` → `$INTERMEDIATE` → `ValueError`).
- A `.env` file (copied from `.env.example`) is the single source of truth, sourced by both Python and the SLURM scripts.
- Roots involved: `INTERMEDIATE`, `PROCESSED`, `TRAINING_TASKS_DIR`, `TASK_DIR`, `FINAL_DATA_DIR`, `OUTPUT_DIR`, …

**MEDS_EIC_AR (upstream, source-verified):** *no* `dotenv`, *no* `.env` file, *no* in-package env read for paths.

- Path roots are passed as **Hydra args only** (`input_dir=`, `intermediate_dir=`, `output_dir=`), backed by whatever shell vars the user chooses to manage themselves.
- Env vars flow **outward, not inward**: `preprocessing/__main__.py` copies `os.environ`, sets `RAW_MEDS_DIR` / `MTD_INPUT_DIR` *from the Hydra args*, and passes that env to the `MEDS_transform-pipeline` subprocess. The MEDS-transforms YAML then reads them via `${oc.env:RAW_MEDS_DIR}`.
- The only genuinely user-tunable env vars are `MIN_SUBJECTS_PER_CODE` / `MIN_EVENTS_PER_SUBJECT`, and those have inline YAML defaults (`${oc.env:MIN_SUBJECTS_PER_CODE,1000}`) — nothing ever "must be set or raise".

So the chain in MEICAR is:

```
user CLI: input_dir=/x  ->  cfg.input_dir  ->  env["RAW_MEDS_DIR"]  ->  subprocess  ->  ${oc.env:RAW_MEDS_DIR} in MEDS-transforms YAML
```

The env var exists only because MEICAR doesn't own the downstream MEDS-transforms config; it's a transient bridge, not a config knob. EveryQuery promoted that internal mechanism into a user-facing dotenv layer.

## Why shift

- **Match upstream conventions** so EveryQuery stays a thin, recognizable extension of MEDS_EIC_AR (easier to track upstream changes, easier for MEDS-ecosystem users).
- **One config mechanism instead of two.** Today paths can come from `.env` *or* Hydra overrides; Hydra is already the config system for everything else (model, datamodule, sampler stages). Folding paths into Hydra removes the parallel dotenv path.
- **Explicit provenance.** `key=value` on the command line / in a config group is self-documenting in run logs; a `.env` read is invisible in the captured Hydra config.

## Open questions / migration considerations

- **Cluster ergonomics.** `.env` is currently the single file sourced by both Python and the SLURM scripts. If we drop the in-package read, the SLURM scripts still need *some* shared source of path vars — likely an `export`-based `env.sh` the scripts `source` and that expands into Hydra args. Need to confirm this doesn't regress the "edit one file to move machines" property.
- **`_resolve_path()` removal.** Decide whether to keep a "must be set" guard (nicer error than an empty `${oc.env:...}` expansion) or lean fully on Hydra's missing-key errors.
- **`${oc.env:VAR,default}` adoption.** For anything we *do* want env-overridable (e.g. threshold knobs), use OmegaConf's `oc.env` resolver with inline defaults instead of `os.environ.get` in Python — matches MEICAR exactly.
- **Backward-compat window.** `.env` is documented across the README and `generate_tasks/README.md`; a transition likely needs a deprecation period where both work.
- **Sequencing.** This touches the same path-resolution code as the in-flight sampler redesign (#202 epic). Probably want to land after that stabilizes to avoid churn.

## Scope (rough)

- `generate_tasks/sample_tasks.py`, `sample_evaluation_tasks.py` — `_resolve_path` / `resolve_training_task_paths` / `load_dotenv`.
- `preprocessing/__main__.py` — already close to MEICAR; check the env-bridge matches.
- `train/`, `predict/`, `evaluate/` entrypoints — any `os.environ`/dotenv reads.
- `.env.example`, README path docs, SLURM scripts.

Not proposing a specific PR yet — this issue is to agree on the target design and sequencing.

gkondas:
## Summary

Addresses #235 — shifts EveryQuery's path/config handling from the `.env` + `load_dotenv()` + env-var-fallback pattern to the upstream MEDS_EIC_AR style, where **path roots are plain Hydra args** and env vars are no longer read inside the package.

Decisions (confirmed with the issue author): **full MEICAR cut** (drop `.env`/`load_dotenv` entirely), implemented **now on `dev-task-sampler`** alongside the #202 sampler work.

## What changed

- **Removed `_resolve_path`** and all `load_dotenv()` calls (`sample_tasks.py`, `sample_evaluation_tasks.py`, `train.py`). `resolve_training_task_paths` now reads `cfg.data_dir` / `cfg.out_dir` directly — these are `???` (mandatory) in the configs, so an unset root raises OmegaConf's `MissingMandatoryValue` instead of silently falling back to an env var.
- **`read_query_codes` takes a `codes_dir` arg** instead of reading `$PROCESSED`; the default query universe is `{codes_dir}/metadata/codes.parquet` (`codes_dir=$PROCESSED`), only consulted when `query_codes`/`codes` is null.
- **Config keys**: sampler/eval `data_dir`/`out_dir` → `???`; `codes_dir` → `null` (optional). Train `tensorized_cohort_dir`/`task_labels_dir` → `???`.
- **Two deliberate env reads kept** (flagged in the diff): train `output_dir: ${oc.env:OUTPUT_DIR}/${run_id:}/` (composes a resolver suffix, so a plain `???` arg doesn't fit) and `entity: ${oc.env:WANDB_ENTITY,null}` (read natively by `wandb`, not our path config).
- **`preprocessing/__main__.py` untouched** — already MEICAR-correct (env flows *outward* to the `MEDS_transform-pipeline` subprocess), as are the `${oc.env:MIN_*,default}` knobs.
- **Dependency/docs**: dropped `python-dotenv`; replaced `.env.example` with an `export`-based `env.example.sh` (sourced by the shell / cluster SLURM scripts, preserving "edit one file to move machines"); updated both READMEs and `redesign-spec.md`.
- **Tests**: dropped the `_resolve_path`/env-fallback matrix; tests now pass paths via cfg/CLI (`codes_dir=`, `data_dir=`, `out_dir=`) and assert `MissingMandatoryValue` on an unset root.

## Migration note

This is a **hard cut** (no `.env` deprecation window), appropriate on the unstable `dev-task-sampler` staging branch. Runtime callers must now `source env.sh` and pass `key=$VAR` Hydra overrides; cluster SLURM scripts (out of repo) need updating to source `env.sh` and append the path args.

## Verification

- `tests/sampler/test_pure_helpers.py`, `test_orchestration.py` — pass (33)
- `tests/test_e2e_foundation.py` — pass (4; exercises the new `codes_dir` arg through the real sampler CLI)
- `tests/test_generate_tasks.py`, `test_generate_evaluation_tasks_cli.py` — pass (20)
- doctests on the three changed modules — pass (16)

https://claude.ai/code/session_01UePA1nWtssNpMQdkwsEf2i


gkondas:
## ✂️ Simplification review (agent 2)

Clean hard-cut PR — most of the diff is deletion, which is exactly right. A few genuine simplifications:

- **`sample_evaluation_tasks.py:431`** — the `codes_cfg is None` ternary is redundant. `read_query_codes` ignores its `codes_dir` arg whenever `codes_or_path` is non-None, so both branches collapse to one call: `codes = read_query_codes(codes_cfg, cfg.get("codes_dir"))`. Drops the local `codes_cfg` too.

- **`sample_tasks.py:635-636` and `sample_evaluation_tasks.py:427-428`** — `Path(str(cfg.x))` double-wraps. For mandatory `???` string keys the value is already a str, so `Path(cfg.x)` suffices. Minor and defensive; only worth it if you want the diff to read as "plain Hydra args, no ceremony."

- **`resolve_training_task_paths` (`sample_tasks.py:617-639`)** — now a 3-line body (two of which are the `Path(str(...))` wraps above) plus a sibling-dir derivation. It is borderline-trivial post-cut, but it still earns its keep: it owns the artifacts-sibling invariant (#7) and has a dedicated unit test. Keeping it is the right call — flagging only so the reviewer knows it was considered, not overlooked.

Everything else (deleted `_resolve_path`, the env-fallback test classes, the `dotenv` dep, the docstring/comment rewrites) is correctly removed with no leftover indirection. No backward-compat cruft survives the hard cut.

gkondas:
## 🔍 Correctness review (agent 1)

One real correctness finding; the rest of the PR (changed `read_query_codes`/`resolve_training_task_paths` signatures, all callers, the `???` config keys, conftest/test updates) checks out.

### `output_dir` validation no longer reports the intended friendly error when `$OUTPUT_DIR` is unset

**`src/every_query/train/configs/config.yaml:4`** + **`src/every_query/train/train.py:246-255`**

The config changed `output_dir: ${oc.env:OUTPUT_DIR,null}/${run_id:}/` → `${oc.env:OUTPUT_DIR}/${run_id:}/` (dropped the `,null` default). But `validate_training_config` was *not* updated to match — it still assumes the old stringified-`None` behavior:

```python
# An unset $OUTPUT_DIR resolves the shipped ``${oc.env:OUTPUT_DIR,null}/${run_id:}/``
# interpolation to the *truthy* string ``'None/<run_id>/'`` ...
output_dir = cfg.get("output_dir")
if not output_dir or str(output_dir).startswith(("None/", "null/")):
    raise ValueError("output_dir is unset (is $OUTPUT_DIR set?). ...")
```

With no `,null` default, resolving `${oc.env:OUTPUT_DIR}` when the env var is unset raises **before** this guard. Verified:

```
cfg.get("output_dir")
-> InterpolationResolutionError: KeyError raised while resolving interpolation:
   "Environment variable 'OUTPUT_DIR' not found"
```

Consequences:
1. The friendly `ValueError("output_dir is unset (is $OUTPUT_DIR set?)...")` is now unreachable for the unset-`OUTPUT_DIR` case; the user instead gets an opaque OmegaConf `InterpolationResolutionError`.
2. The `str(output_dir).startswith(("None/", "null/"))` branch and its multi-line comment (lines 246-249) are now dead code documenting behavior that can no longer occur.

It still fails loudly (no silent wrong path), so this is a UX/maintenance regression rather than a data-corruption bug — but the validator is now internally inconsistent with the config it validates.

**Fix (pick one):**
- Restore the default in config.yaml: `output_dir: ${oc.env:OUTPUT_DIR,null}/${run_id:}/` (keeps the existing validator correct and the friendly message intact); or
- Keep `${oc.env:OUTPUT_DIR}` and rewrite the validator to catch the resolution failure, e.g. wrap `cfg.get("output_dir")` in a try/except re-raising the friendly `ValueError`, and delete the now-dead `None/`/`null/` guard + comment.

### Minor (non-blocking, not correctness)
- `src/every_query/generate_tasks/sample_evaluation_tasks.py:413-416` — `main()` docstring still says "Path fallback mirrors `sample_tasks`: `cfg.data_dir` -> `$INTERMEDIATE`, `cfg.codes_dir` -> `$PROCESSED`, `cfg.out_dir` -> `$TASK_DIR`", which no longer matches the now-required-arg behavior. Stale comment only.
- `train.py:239,243,254` error strings still say "set $... in your .env"; harmless but `.env` is gone.

gkondas:
## 📄 Docs & migration review (agent 3)

Reviewed for documentation accuracy and migration completeness. The PR's own edited files (root `README.md`, `generate_tasks/README.md`, `redesign-spec.md`, both sampler config YAMLs, `env.example.sh`, `pyproject.toml`) are internally consistent and accurate. But the migration is **incomplete**: it left stale `.env`/`load_dotenv` text in files it did not touch.

### Stale references that must be fixed

**`src/every_query/train/README.md` (not edited by this PR)** — still documents the old env-fallback convention that #235 removed:
- `:15` — `task_labels_dir ... default ${oc.env:TRAINING_TASKS_DIR}`. After this PR `task_labels_dir` is `???` (mandatory, no env default). Stale.
- `:38` — "...the default for `datamodule.config.task_labels_dir` — no override needed when `.env` is set." Directly contradicts the new contract (must pass it; no `.env`). Stale.

**`src/every_query/train/train.py` (PR edited `_init_env` + the wandb message, but not the path-validation messages/comments)** — error strings still tell users to use `.env`, and one comment describes a config default this PR removed:
- `:239` — `"... or set ${env_var} in your .env."` should be "export `$VAR`" (matches the wandb message you already fixed at `:260`).
- `:254` — `"... or set $OUTPUT_DIR in your .env."` same.
- `:246-249` — comment describes the `${oc.env:OUTPUT_DIR,null}` interpolation and the resulting literal `'None/<run_id>/'` path. But `config.yaml:4` was changed in this PR to drop the `,null` default, so an unset `OUTPUT_DIR` now raises an interpolation-resolution error instead of stringifying to `None/...`. The comment (and the `startswith(("None/","null/"))` guard it explains) is now stale/misleading.

### Scoping inaccuracy (softer)

- `README.md:262-264` asserts "The two genuine env reads that remain: `OUTPUT_DIR` and `WANDB_ENTITY`." At the repo level that undercounts: `preprocessing/configs/_data.yaml` & `_reshard_data.yaml` read `RAW_MEDS_DIR`, `MTD_INPUT_DIR`, `MIN_SUBJECTS_PER_CODE`, `MIN_EVENTS_PER_SUBJECT`; `predict/external_tasks/configs/*.yaml` read `TASK_DIR`, `ACES_SHARDS_DIR`, `OUTPUT_DIR`. If the sentence is meant to be scoped to train + samplers, please say so explicitly.

### Migration gap: dropped var

- The old `.env.example` documented `ACES_SHARDS_DIR` (commented, for the `aces_to_eq` pipeline). `env.example.sh` dropped it, yet `predict/external_tasks/configs/aces_to_eq.yaml:16` still reads `${oc.env:ACES_SHARDS_DIR}`. Add it back (commented) for parity. The preprocessing env vars (`RAW_MEDS_DIR`, `MTD_INPUT_DIR`, ...) were absent from both files — pre-existing gap, not a regression.

### Verified good
- No leftover `.env.example`/`load_dotenv`/`dotenv` references in the edited docs or sampler code; `_resolve_path` and its tests were removed cleanly; `dotenv` dropped from `pyproject.toml`.
- `env.example.sh` covers every var the samplers + train need (`INTERMEDIATE`/`PROCESSED`/`FINAL_DATA_DIR`/`TASK_DIR`/`TRAINING_TASKS_DIR`/`OUTPUT_DIR`/`WANDB_*`), with `key=` annotations.
- The "source one file (incl. SLURM scripts)" property is still documented and now accurate (shell-owned `export`s vs. the old dotenv-sourced file).
- README CLI examples now pass the required `data_dir`/`out_dir`/`codes_dir` args; `conftest.py` fixture updated to pass `codes_dir=` instead of `env={"PROCESSED": ...}`.

gkondas:
## 🧭 API & upstream-consistency review (agent 6)

Scope: internal API consistency + fidelity to the #235 MEICAR goal (path roots = plain Hydra args; env vars only as an outward subprocess-plumbing detail). Verified against the checked-out branch, not just the diff.

**Verdict: the design is principled and faithful to #235.** The two findings below are minor/pre-existing, not blockers.

### What's correct and consistent

- **`preprocessing/__main__.py` genuinely matches the MEICAR outward-bridge pattern** the issue describes — this is not a hollow claim. `main()` reads `input_dir`/`intermediate_dir`/`output_dir` as plain Hydra args (all `???` in `configs/_process_data.yaml`), then *derives* `env["RAW_MEDS_DIR"]`/`env["MTD_INPUT_DIR"]` from those args and passes the copied env outward to the `MEDS_transform-pipeline` subprocess, which reads them via `${oc.env:RAW_MEDS_DIR}` in `_data.yaml`/`_reshard_data.yaml`. Env flows out, never in. Correctly left untouched by this PR.
- **Sampler path keys are handled consistently** between the two entrypoints: `sample_tasks.py`'s `resolve_training_task_paths` does `Path(str(cfg.data_dir))` / `Path(str(cfg.out_dir))`, and `sample_evaluation_tasks.py` `main()` does the same — both `data_dir`/`out_dir` are `???` and both raise `MissingMandatoryValue` on access when unset. `codes_dir` is `null` (optional) in both, consulted only when `query_codes`/`codes` is null. Symmetric.
- **The `output_dir` vs `???` split in the train config is defensible, not an inconsistency.** `output_dir: ${oc.env:OUTPUT_DIR,null}/${run_id:}/` composes a `/${run_id:}/` suffix at config-compose time, so a bare `???` arg can't express it; `tensorized_cohort_dir`/`task_labels_dir` are plain leaf paths and are correctly `???`. The `,null` default is load-bearing: I confirmed dropping it would raise `InterpolationResolutionError` before `validate_training_config`'s friendly "output_dir is unset" guard could fire — follow-up commit `c10d4b6` restoring `,null` is the right call. `WANDB_ENTITY` staying `${oc.env:...,null}` is also principled: it's read natively by `wandb`, not by EveryQuery's path layer. Both kept reads are documented in the README "Paths & environment" table.
- **The `???` → friendly-error path actually works.** I verified `ds_cfg.get(node)` on a `???` node returns `None` (not a raise) in this OmegaConf version, so `validate_training_config`'s `ValueError` ("…typically =$FINAL_DATA_DIR") stays reachable for the train keys. The samplers deliberately take the rawer `MissingMandatoryValue` path via direct `cfg.data_dir` access — a defensible difference (train has a dedicated validation layer; the samplers don't).

### Minor findings (non-blocking)

1. **Pre-existing naming divergence from upstream MEICAR, now also internally inconsistent.** Upstream MEICAR and this repo's own `preprocessing` config use `input_dir`/`intermediate_dir`/`output_dir`; the samplers use `data_dir`/`out_dir`. So `EQ_process_data` takes `output_dir=` while the samplers take `out_dir=` for an analogous role. This PR did not introduce it (it only swapped the resolution mechanism, not the key names) and #235 explicitly scoped naming out, but it's worth a tracking note: the stated goal is "stay a thin, recognizable extension of MEDS_EIC_AR," and `data_dir`/`out_dir` is the one spot that still reads non-MEICAR. The `query_codes` (training) vs `codes` (eval) key asymmetry for the same `read_query_codes(codes_or_path, codes_dir)` positional is likewise pre-existing and cosmetic.

2. **One stale `.env` reference survives in source.** `src/every_query/predict/external_tasks/configs/aces_to_eq.yaml:15` still reads `# ACES_SHARDS_DIR lives outside TASK_DIR/OUTPUT_DIR — set it in .env if you use this pipeline.` Commit `c10d4b6`'s message says it "scoped the README note to include preprocessing/aces," but this in-config comment now points at a `.env` file the PR deleted. Should say `env.sh` (or just "export it"). Cheap one-line fix.

Everything else (dotenv dependency drop, `env.example.sh` replacing `.env.example`, test migration to cfg-supplied paths + `MissingMandatoryValue` assertions, no orphaned `os` import in the samplers) is internally consistent.

gkondas:
## ⚠️ Edge-case & error-UX review (agent 5)

Scope: edge cases + error-message UX only. One real regression (with two faces), one place the error surfaces late, and a couple of things that are robust. Repro is `uv run python` against the actual helpers on this branch.

### 🔴 1. Empty env-var expansion → `Path("None")`, silently (the important one)

The `.env`-era contract treated an *unset* path root as an error. With `data_dir: ???`, the intended guard is `MissingMandatoryValue`. But the common shell usage is `data_dir=$INTERMEDIATE`, and if `$INTERMEDIATE` is unset the shell expands that to `data_dir=` (empty). Hydra parses `data_dir=` as **`None`**, which *overrides* the `???` sentinel — so the key is now present-and-null, and `MissingMandatoryValue` never fires:

```python
base = OmegaConf.create({'data_dir': '???', 'out_dir': '???'})
cfg  = OmegaConf.merge(base, OmegaConf.from_dotlist(['data_dir=', 'out_dir=']))
resolve_training_task_paths(cfg)
# -> (PosixPath('None'), PosixPath('None'), PosixPath('None_artifacts'))
```

`Path(str(cfg.data_dir))` stringifies `None` into a **literal directory named `None`**. The run then proceeds into Stage 0 and only dies there with:

```
FileNotFoundError: No shards found under None/data/train; expected {i}.parquet files.
```

Why it's a problem: (a) the failure is silent at the resolution boundary and confusingly blames a path `None/data/train` the user never typed, never mentioning the real cause (empty env var); (b) it's a **regression** — the old `_resolve_path(None, "INTERMEDIATE", "data_dir")` hit the env-var branch with an unset `$INTERMEDIATE` and raised up front with "data_dir must be set… export $INTERMEDIATE". This is precisely the case the old code's truthy `if env_value:` check was guarding. Same code path in both `sample_tasks.py::resolve_training_task_paths` and `sample_evaluation_tasks.py::main` (`Path(str(cfg.data_dir))` / `Path(str(cfg.out_dir))`).

### 🟠 2. Literal empty string → `Path(".")`

A sibling of the above, for the rare `data_dir=""` (quoted literal empty):

```python
cfg = OmegaConf.create({'data_dir': '', 'out_dir': ''})
resolve_training_task_paths(cfg)
# -> (PosixPath('.'), PosixPath('.'), PosixPath('_artifacts'))
```

`out_dir="."` means the sampler would write its final dataset and `_artifacts` tree into the **current working directory** — no guard at all. The old `_resolve_path` env branch rejected empties; this silently accepts them.

### ✅ Fix for 1 & 2

Validate the two roots explicitly in `resolve_training_task_paths` (and the eval `main`), up front, instead of relying solely on `???`. e.g.:

```python
def _require_dir_arg(cfg, key):
    v = cfg.get(key)
    if v is None or str(v).strip() == "":
        raise ValueError(
            f"{key} is unset/empty. Pass {key}=<path> (typically a shell-expanded "
            f"$INTERMEDIATE/$TRAINING_TASKS_DIR — check that var is actually exported)."
        )
    return Path(str(v))
```

This restores the clear up-front error, catches both the `None` (empty-env) and `""` (literal) faces, and names the likely culprit (unexported env var). Keeping it in the resolve helper means it fires at the resolution boundary, before any Stage 0 shard scanning.

### ✅ 3. `read_query_codes` null/empty handling is robust — no regression

These are all good and need no change:

- `read_query_codes(None, None)` → `ValueError: query_codes is null and no codes_dir was given; pass query_codes=... or codes_dir=...` — clear.
- `read_query_codes(None, "")` → same clear `ValueError` (the `if not codes_dir:` truthy check catches empty string — so the empty-env case is handled correctly *here*, unlike the path roots above). `codes_dir=$PROCESSED` with unset `$PROCESSED` → `codes_dir=` → `None` → caught. 👍
- `read_query_codes(None, "/missing/dir")` and an existing dir lacking `metadata/codes.parquet` → `FileNotFoundError: ... /metadata/codes.parquet`. Decent (names the exact expected file).

The asymmetry is the tell: `read_query_codes` guards `not codes_dir`, but `resolve_training_task_paths` does a bare `Path(str(...))`. Applying the same truthy guard to the path roots closes the gap.

### Timing note

`resolve_training_task_paths` is called first in `run()` (good — `MissingMandatoryValue` for a *truly* absent root surfaces before any work). But `read_query_codes` runs at Stage 1 via `QueryDistribution.from_config`, *after* Stage 0 has scanned every shard. A bad `codes_dir` therefore wastes the full Stage 0 pass before failing. Minor, but if cheap, resolving/validating `codes_dir` up front in `run()` next to the path roots would fail faster.

gkondas:
## 🧪 Test-quality review (agent 4)

Scope: test quality/coverage of the MEICAR path-handling cut only. I ran `tests/sampler/test_pure_helpers.py` + `tests/sampler/test_orchestration.py` (33 passed) and verified the OmegaConf `???` / `.get()` semantics directly. Net assessment: **the migration is well-tested on the training + train-validation sides; the gaps are all on the evaluation CLI and at the CLI boundary for the new error paths.**

### Verified solid (not vacuous)
- `test_pure_helpers.py:946` `test_missing_required_root_raises` — confirmed real: `OmegaConf.create({"data_dir": "???"})` genuinely produces MISSING and accessing it raises `MissingMandatoryValue`. Not a literal-string false pass.
- `test_pure_helpers.py:968` `test_required_keys_present_and_typed` — `OmegaConf.is_missing(...)` is the right shape assertion for `???`.
- `test_pure_helpers.py:827/841` `read_query_codes(None, codes_dir)` / `(None, None)` — both exercise the real branch.
- Train side: `validate_training_config`'s friendly `ValueError` guard is still reachable under the shipped `???` config — verified `ds_cfg.get(node)` returns `None` for a MISSING node (it swallows MISSING), so the `if not value` guard fires before any raw `MissingMandatoryValue`. The `None`-based cases in `test_env_validation.py:57-90` therefore faithfully cover the new `???` defaults. Dropping the env-fallback matrix (`TestResolvePath`) is correct — that behavior no longer exists.

### Gaps worth closing

**1. Eval CLI never exercises the default-query-universe branch (the exact line this PR changed).**
`sample_evaluation_tasks.py:444` collapses the old ternary into one `read_query_codes(cfg.get("codes"), cfg.get("codes_dir"))`. But all three eval CLI tests (`test_generate_evaluation_tasks_cli.py:53,109,164`) pass `codes=[HR,TEMP]` explicitly, so `codes_dir` is always ignored and the `codes`-null → `{codes_dir}/metadata/codes.parquet` path is never run through the real eval CLI. (The *training* CLI does cover this — `conftest.py:228` leaves `query_codes` unset and passes `codes_dir=`, so that side is fine.) The collapsed eval call is the new code and is untested end-to-end.
Suggested: a test that drops `codes=` and relies on `codes_dir=$eq_preprocessed_dataset`, asserting the output is non-trivial and queries come from the metadata universe — the analogue of the training fixture's default-universe run.

**2. No CLI-level test that an unset required root surfaces a clean error.**
`MissingMandatoryValue` is only proven at the `resolve_training_task_paths` unit level. There's no test that running `EQ_generate_training_tasks`/`EQ_generate_evaluation_tasks` with `data_dir`/`out_dir` omitted fails loudly (and doesn't silently fall back to some ambient value). Given the whole point of #235 is "unset root must raise, not fall back to env", a CLI-boundary regression guard is the one that actually protects the contract.
Suggested: one `run_and_check`-style test asserting a nonzero exit / `MissingMandatoryValue` in stderr when `out_dir=` is omitted.

**3. `read_query_codes(None, None)` is only unit-tested, never via the real CLI.**
Minor given #1/#2, but if you add the eval default-universe test, also worth one case where both `codes`/`query_codes` and `codes_dir` are null through the CLI, asserting the `"query_codes is null"` ValueError reaches the user (not an obscure traceback). Currently `test_read_query_codes_requires_codes_dir_when_null` proves the function raises, but nothing proves the CLI presents it.

**4. (Minor) Train `???` keys tested only via `None`, not the literal shipped sentinel.**
`test_env_validation.py` builds configs with `None`; the shipped config is `???`. Equivalent at runtime (verified above), but a single case loading the real `config.yaml` and asserting `OmegaConf.is_missing(..., "tensorized_cohort_dir")` would lock the config-shape change the way `test_required_keys_present_and_typed` does for the sampler. Nice-to-have, not blocking.

Nothing here is a correctness bug — items 1 and 2 are the substantive coverage gaps (new code / core #235 contract untested at the CLI boundary).

gkondas:
## 🔁 Correctness re-review (agent 7)

Reviewed at HEAD `5ebc8b0` (base `dev-task-sampler`), focused on the newly-added `_require_path_arg` guard and the broader env-var removal. **No correctness issues found — the current state is clean.**

### `_require_path_arg` guard — correct
- `cfg.get(name)` reliably returns `None` for a `???` MISSING key. Verified empirically on the pinned **omegaconf 2.3.0**: `cfg.get('data_dir')` → `None` while `OmegaConf.is_missing` is `True`, and a present sibling key returns its value. So the guard never reaches the `MissingMandatoryValue`-raising attribute access; it returns a clean `ValueError` for both the truly-unset (`???`) and empty-override (`key=` from an unexported `$VAR`) cases, exactly as the docstring claims.
- `None`/`''`/`'   '` (whitespace-only) all correctly raise; `str(value).strip()` → empty → `not s` → raise. Verified.
- **No valid path is wrongly rejected:** relative paths (`rel/path`), `.`, and paths with surrounding whitespace all pass. `.strip()` only trims leading/trailing whitespace (which `Path` would discard anyway) and never mangles interior characters — `'  /a/b  '` → `Path('/a/b')`. A path that is *only* whitespace is not a legitimate path, so rejecting it is correct.

### Internal consistency — verified
- Both call sites updated: `resolve_training_task_paths` (sample_tasks.py:654-655) and eval `main` (sample_evaluation_tasks.py:428-429). The import swap `_resolve_path`→`_require_path_arg` in sample_evaluation_tasks.py is consistent.
- `resolve_training_task_paths` dropped its `cfg=None` default, but its only non-test caller (`run()`, sample_tasks.py:1391) always passes `cfg`; the removed `resolve_training_task_paths()` no-arg test was deleted in the same diff. Consistent.
- `read_query_codes` two-arg signature: explicit list/path branch ignores `codes_dir`; `None`+`None` raises the new clear message. `from_config` (sample_tasks.py:152) and eval `main` (codes.parquet:433) both pass `cfg.get('codes_dir')`. Verified empirically.
- **No remaining reads of removed env vars in executable code.** Grep for `_resolve_path`/`load_dotenv`/`os.environ.get('PROCESSED'|'INTERMEDIATE'|'TRAINING_TASKS_DIR'|'TASK_DIR')` across `src/` hits only a doc-comment reference and stale PKG-INFO (generated artifact) — no live code. `train.py:_init_env` correctly dropped `load_dotenv` while keeping SLURM thread-count logic; `validate_training_config` remains reachable (called at train.py:278).
- Eval config keys confirmed (`data_dir/out_dir: ???`, `codes/codes_dir: null`) so the eval default-universe path raises cleanly when `codes_dir` is unset.
- The genuine remaining `${oc.env:...}` reads (`OUTPUT_DIR`, `WANDB_ENTITY` in train; `TASK_DIR`/`ACES_SHARDS_DIR` in aces_to_eq) are intentional and documented.

### Tests
`tests/sampler/test_pure_helpers.py` — 25 passed (includes the new `test_missing_required_root_raises` and `test_empty_override_raises_not_silent_none` covering the `None`/`''`/`'   '` matrix).

LGTM.

gkondas:
## ✅ Merge-readiness review (agent 8)

Reviewed current HEAD `5ebc8b0` (3 commits) against the #235 "Scope (rough)" list. **Verdict: clean — ready to merge.** No blocking issues found.

### Scope coverage (#235)
| Scope item | Status |
| --- | --- |
| `sample_tasks.py` — drop `_resolve_path`/`load_dotenv` | ✅ removed; `resolve_training_task_paths(cfg)` reads `cfg.data_dir`/`cfg.out_dir` via new `_require_path_arg` |
| `sample_evaluation_tasks.py` | ✅ `load_dotenv`/`_resolve_path` gone; single `read_query_codes(codes, codes_dir)` call |
| `preprocessing/__main__.py` — env-bridge check | ✅ correctly **untouched** (0 files in diff under `preprocessing/`); outward `${oc.env:...}` bridge already MEICAR-correct, matching the PR description |
| `train`/`predict`/`evaluate` entrypoints | ✅ `train.py` `_init_env()` drops `load_dotenv` (keeps `os.environ` thread-count read); `predict/` & `evaluate/` have no dotenv/path-env reads |
| `.env.example` | ✅ replaced with `export`-based `env.example.sh`; `.gitignore` now ignores `env.sh` |
| README path docs | ✅ root README + `generate_tasks/README.md` + `train/README.md` + `redesign-spec.md` all updated |
| SLURM-script guidance | ✅ documented (scripts live out of repo): `env.example.sh` header + README "Paths & environment" both state SLURM scripts should `source env.sh` and append `key=$VAR` args |

No scope item left unaddressed.

### Loose-end hunt (all clear)
- **No orphaned refs.** Every surviving `.env`/`oc.env`/`_resolve_path` hit is intentional: doc references to #235, the legit preprocessing/wandb/aces `${oc.env:...}` interpolations, and one historical mention of `_resolve_path` *inside* the new `_require_path_arg` docstring (`sample_tasks.py:611`) — explanatory, not dead code. No `dotenv`/`ensure_env` left anywhere in tracked files (egg-info excluded).
- **No dead imports.** `import os` in `sample_tasks.py` and `train.py` still used (`os.replace`, `POLARS_MAX_THREADS`, `os.cpu_count`).
- **`uv.lock` correct.** Both `dotenv` and the transitive `python-dotenv` package stanzas removed; `everyquery` deps + `requires-dist` no longer list `dotenv`; 0 remaining refs. `pyproject.toml` dep dropped too.
- **PR description accurate** — does not overstate modified files (verified 20-file diff matches: preprocessing genuinely untouched, two deliberate train env reads `OUTPUT_DIR`/`WANDB_ENTITY` correctly flagged).
- **Nice design touch:** `_require_path_arg` guards the empty-override footgun (`data_dir=$VAR` with an unexported var → empty string overriding `???` → silent `Path("None")`). Covered by `test_empty_override_raises_not_silent_none`.

### Commit history
Coherent 3-commit arc: initial cut → round-1 review (restore `output_dir` null default + stale docs) → round-2 review (empty/unset guard + stale-doc fixes). Reads cleanly.

### Verification
`uv run pytest tests/sampler/test_pure_helpers.py tests/sampler/test_orchestration.py` → **34 passed**. Test suite correctly migrated off the env-fallback matrix to cfg/CLI-driven assertions (`MissingMandatoryValue`/`ValueError` on unset roots).

### Nice-to-haves (non-blocking)
- None material. The out-of-repo SLURM scripts still need the actual `source env.sh` + `key=$VAR` edit, but that is correctly out of scope for this repo PR and is documented for whoever owns those scripts.

gkondas:
## 🤖 Multi-agent review summary

Three agents reviewed PR #245 at `1b3e033` along separate axes (runtime correctness, tests/config consistency, simplicity/design). Consolidated verdict below.

### ✅ Verdict: good to merge

**As committed, this PR is green and sound.** `uv run pytest tests/sampler/ tests/test_generate_tasks.py` → **116 passed, 0 failed, 0 skipped** (incl. the 3 integration tests that run the real sampler CLI against `simple_static_MEDS`). All of the PR's "Verification" claims check out.

**Correctness** — Runtime path flow traced end-to-end. `_require_path_arg` correctly rejects both the unset-`???` and unexported-`$VAR`→empty cases (Hydra parses the latter as `None`, which would otherwise silently override the sentinel and yield `Path("None")`). No leftover `load_dotenv`/`_resolve_path`/`os.environ` reads. The two retained `oc.env` reads (`OUTPUT_DIR` needs the `${run_id:}` resolver suffix; `WANDB_ENTITY` is read natively by `wandb`) are correctly scoped and documented.

**Tests/config** — `python-dotenv` fully removed (pyproject + uv.lock), `.env.example`→`env.example.sh` consistent with the keys the code reads, `.gitignore` covers both. New code paths (`MissingMandatoryValue`, no-env-fallback) are covered. The lone `load_dotenv` mention in README is the intentional "the package does NOT read .env" note, not stale guidance.

**Simplicity** — Genuine net simplification (−332/+230): removes a dependency, 3 `load_dotenv()` sites, and the `_resolve_path` override>env>raise matrix + its test class. Complexity removed, not relocated. `codes_dir=null` (optional) vs `data_dir/out_dir=???` (mandatory) is a correct conditional-need distinction. No new abstractions or dead flexibility.

No inline comments posted — nothing block-worthy was found in the committed diff.

### ⚠️ One caveat (not in this PR, but relevant before pushing)
There are **uncommitted local working-tree edits** — a half-finished follow-up that switches the training config `query_codes: null → ???` and drops `codes_dir`. It's currently inconsistent: `tests/test_generate_tasks.py:187` was updated to `query_codes=` but `conftest.py:228` and `tests/test_generate_tasks.py:243` still pass the removed `codes_dir=` key. **Committing/pushing that as-is would turn the branch red** (`Key 'codes_dir' is not in struct`). Fix those two lines before pushing the follow-up. The committed PR is unaffected.

PR #252, opened 2026-06-24
Drop custom run_id; use Hydra run/sweep dir for EQ_train output uniqueness (#251)

### LINKED ISSUE #251: EQ_train: drop custom run_id, leverage Hydra run/sweep dir for output_dir uniqueness
## Summary

`EQ_train`'s `output_dir` currently relies on a custom timestamp `run_id` resolver for per-run uniqueness while *ignoring* Hydra's built-in run directory. We should drop `run_id` and lean on Hydra's native `run.dir` / `sweep.dir` machinery, which gives us clean single-run paths **and** hyperparameter sweeps for free.

## How it works today

`src/every_query/train/configs/config.yaml`:
```yaml
output_dir: ${oc.env:TRAINING_OUTPUT_DIR,null}/${run_id:}/
...
trainer:
  default_root_dir: ${output_dir}
  logger:
    name: "${run_id:}"
```

`src/every_query/train/train.py`:
```python
@OmegaConfResolver(replace=True)
def run_id():
    global _RUN_ID
    if _RUN_ID is None:
        _RUN_ID = datetime.now(UTC).strftime("%Y-%m-%d/%H-%M-%S")
    return _RUN_ID
```

So uniqueness = `$TRAINING_OUTPUT_DIR/<date>/<time>/`, and `default_root_dir` (checkpoints, logger save_dir, config dumps, `best_model.ckpt`) all hang off that.

### The problem

This runs **two parallel uniqueness systems**:

| Mechanism | Path | Contents | Used? |
|---|---|---|---|
| Custom `run_id` resolver | `<base>/<date>/<time>/` | checkpoints, configs, `best_model.ckpt` | ✅ our real output |
| Hydra's built-in run dir | `outputs/<date>/<time>/` | `.hydra/` metadata + log | ❌ orphaned clutter |

Because the code never reads `HydraConfig.get().runtime.output_dir`, every invocation drops a stray `outputs/<date>/<time>/` folder next to the launch dir, separate from where artifacts actually go. The `run_id` resolver is a hand-rolled duplicate of Hydra's built-in `${now:}`.

It also has no sweep story: launching N hyperparameter configs has no native per-run folder naming.

## Proposed: leverage Hydra's run dir properly

Model on MEDS_EIC_AR (`_pretrain.yaml`), which makes `output_dir` a required base and redirects Hydra's run dir into it. Extend it with sweep support.

**`config.yaml`:**
```yaml
output_dir: ???   # required: user supplies the base path

hydra:
  run:
    dir: ${output_dir}                        # single run → artifacts directly in output_dir
  sweep:
    dir: ${output_dir}                        # sweep base
    subdir: ${hydra.job.override_dirname}     # one named subdir per job (e.g. lr=1e-4,weight_decay=0.05)

trainer:
  default_root_dir: ${hydra:runtime.output_dir}   # single source of truth: run.dir OR sweep.dir/subdir
  logger:
    name: null   # drop ${run_id:}; let wandb auto-name (or use ${hydra.job.override_dirname})
```

**`train.py`:**
- Delete the `run_id` resolver (and `_RUN_ID`).
- Repoint the run-folder reads (overwrite/resume logic, `cfg_path`, `best_model.ckpt`) from `cfg.output_dir` to the resolved per-job dir, e.g. `Path(cfg.trainer.default_root_dir)` (already `${hydra:runtime.output_dir}`). Otherwise sweep jobs would `rmtree`/write to the shared base and collide.
- Keep the validate guard requiring `cfg.output_dir` (the base) to be set.

### Result

```bash
EQ_train output_dir=/runs/exp1 ...                                   # → /runs/exp1/
EQ_train -m output_dir=/runs/sweep lightning_module.optimizer.lr=1e-5,1e-4,1e-3   # → /runs/sweep/lr=1e-5/ ...
```

- `output_dir` becomes a required, user-supplied base.
- Uniqueness comes from Hydra (`runtime.output_dir`): single = base, sweep = base/param-subdir.
- No more orphaned `outputs/` dirs; `run_id` resolver deleted.

## Follow-up (out of scope here)

`-m` runs jobs **sequentially** in one process. Fanning a sweep across SLURM nodes needs the `hydra-submitit-launcher` plugin — separate issue. This change is the prerequisite.

gkondas:
Closes #251.

## Problem

`EQ_train` ran **two parallel uniqueness systems**: a hand-rolled `run_id` OmegaConf resolver stamped `<date>/<time>` into `output_dir` (where artifacts actually land), while Hydra's own run dir was never read — so every invocation dropped a stray orphaned `outputs/<date>/<time>/` folder (just `.hydra/` metadata) next to the launch dir. The resolver also gave sweeps no per-run folder story.

## Change

Delete the custom resolver and lean on Hydra's native `run.dir` / `sweep.dir`:

- **`config.yaml`**: `output_dir: ???` (required base the user supplies); a `hydra` block routes `run.dir`/`sweep.dir` to `${output_dir}/<date>/<time>` with `sweep.subdir = ${hydra.job.num}_${hydra.job.override_dirname}` (the `job.num` prefix guarantees a unique subdir even if two jobs share an override set); `trainer.default_root_dir` reads `${hydra:runtime.output_dir}` as the single source of truth; wandb uses standard auto-naming (`name: null`) and the resolved `run_dir` is logged as a hyperparam so the UI↔folder link is preserved.
- **`train.py`**: remove the `run_id` resolver, `_RUN_ID`, and the dead `datetime` import; read the per-run/per-job dir from `cfg.trainer.default_root_dir` (not the shared base, so sweep jobs don't collide); simplify the `output_dir` guard.
- Delete `tests/test_run_id.py` and the obsolete `None/<run_id>` env-validation test.
- Update root + train READMEs and the per-code eval config for the new layout.

`${now:}` is fixed once per launch, so `run.dir` and `sweep.dir` share one timestamp — the **same `output_dir` can be reused across runs and sweeps with no collisions**.

```bash
EQ_train output_dir=/runs/exp1 ...                      # → /runs/exp1/<date>/<time>/
EQ_train -m output_dir=/runs/sweep ...optimizer.lr=1e-5,1e-4   # → /runs/sweep/<date>/<time>/lr=1e-05/ ...
```

## Verification

Locally (WSL, no data/GPU): `test_env_validation` (8) + `train.py` doctests (4) pass; a real invocation creates `/runs/.../<date>/<time>/` (confirming `run.dir`/`default_root_dir` resolve), and a missing `output_dir` errors cleanly with `Missing mandatory value: output_dir`. Integration/resume + sweep suites run cluster-side (they use the self-contained `_demo_train.yaml`, which stays B-style so `default_root_dir` resolves to `output_dir` unchanged).

## Follow-up (out of scope)

`-m` runs sweep jobs sequentially in one process. Fanning across SLURM nodes needs the `hydra-submitit-launcher` plugin — separate issue.


gkondas:
## 🧹 Design & test review (automated)

**Test result:** `uv run pytest tests/test_env_validation.py tests/test_train_hydra_dirs.py --doctest-modules src/every_query/train/train.py -q` → **14 passed** (28s, WSL, no data/GPU). The two subprocess `test_train_hydra_dirs.py` tests pass because they fail-fast on bogus data dirs *after* Hydra has materialized the run/sweep dir — the dir-creation assertions still exercise the real `config.yaml` hydra block. 👍

Overall this is a solid simplification: deleting the custom `run_id` resolver + singleton global + its 3 tests in favor of Hydra-native `run.dir`/`sweep.dir` is exactly the right "delete code, lean on the framework" move. A few things worth a look:

- **MEDIUM** — `src/every_query/train/configs/config.yaml:107` & `_demo_train.yaml:73`: **config drift.** Real config now uses `default_root_dir: ${hydra:runtime.output_dir}` (timestamped/subdir-aware), but `_demo_train.yaml` still hardwires `default_root_dir: ${output_dir}` with **no hydra block**. So the integration suites never exercise the new uniqueness machinery — which is precisely why `test_train_hydra_dirs.py` had to be added as a separate subprocess harness. That's an acceptable workaround, but the two configs will keep diverging silently; consider a shared `hydra:` defaults group or a comment in `_demo_train.yaml` flagging the intentional divergence.

- **LOW** — `train.py:244` (`validate_training_config`): the new `trainer.default_root_dir` guard adds little for the *real* Hydra path — `${hydra:runtime.output_dir}` always resolves to a non-empty string, so it can only fire for hand-built configs (the tests). The stated rationale ("a stray `default_root_dir=` override can't pass while artifacts land elsewhere") doesn't really hold: an override to a *valid-but-wrong* dir still passes the `if not ...` truthiness check. It's a cheap guard, but it's closer to defensive boilerplate than a real gate. Fine to keep, just don't oversell it in the comment.

- **LOW** — `train.py:281` (`run_dir.is_file()` guard): for a normal launch Hydra has *already created* `run_dir` before `main()` runs, so this can only trip if someone points `output_dir`/`hydra.run.dir` at an existing file path — a genuine but very narrow case. Redundant-ish with the `os.makedirs` that follows, but harmless and cheap. Keep.

- **LOW** — `config.yaml:23-26` `override_dirname.exclude_keys`: covers the three always-present path args. Watch for two pitfalls not addressed here: (1) `do_overwrite`/`do_resume`/`seed` are *not* excluded, so any sweep that varies them (or that a user passes on the CLI) leaks into the subdir name — usually fine; (2) float overrides like `lr=1e-5` keep their `=`/`.`/`e-` chars, which is OK for one key but a multi-axis sweep produces long, filesystem-marginal dir names. The `${hydra.job.num}_` prefix is the right safety net for uniqueness; just be aware override_dirname names can still get unwieldy. No change required.

- **NIT** — `config.yaml:13-16`: `run.dir` and `sweep.dir` duplicate the same `${output_dir}/${now:...}` string. The comment correctly notes `${now:}` is fixed per launch so they stay consistent, but the duplication is a future drift hazard. Could hoist to a local interpolation var, though OmegaConf makes that awkward; leaving as-is is defensible.

- **NIT** — test coverage gap: `test_train_hydra_dirs.py` asserts the `<date>`/`<time>` dir *lengths* and that swept subdirs contain `lr=` and exclude path keys — but it does **not** assert the `${hydra.job.num}_` prefix that the config comment calls the uniqueness guarantee (e.g. `subdirs` aren't checked for a leading `0_`/`1_`). Adding `assert any(n.startswith("0_") for n in subdirs)` would cover the one piece of the subdir template that's currently untested. The collision case the prefix defends against (two jobs with identical override sets) also isn't tested, but that's harder to construct and lower value.

- **NIT** — `output_dir: ???` vs the repo's dotenv/env-var convention: this intentionally drops the `${oc.env:TRAINING_OUTPUT_DIR,null}` fallback in favor of a required arg, consistent with the #235 direction for the other path args (`tensorized_cohort_dir`/`task_labels_dir` are already `???`). READMEs + `env.example.sh` were updated coherently. Good, no action — just flagging that SLURM scripts now *must* pass `output_dir=$TRAINING_OUTPUT_DIR` explicitly; double-check the cluster launch scripts were updated to match (not in this diff).

- **GOOD** — `resume_check.py:25,29` already allow-lists both `output_dir` and `trainer.default_root_dir` in `ALLOWED_DIFFERENCE_KEYS`, so the new reliance on `default_root_dir` won't trip the resume drift check. README resume guidance (`hydra.run.dir=<...> do_resume=True`) is the correct Hydra incantation. Eval configs (`model_run_dir`/`model_path`) correctly drop the now-gone `/outputs/` segment.

Net: ship-able. The MEDIUM (demo config divergence) is the only thing I'd push on; the rest are NITs/keep-as-is.

gkondas:
## 🔎 Correctness review (automated)

Verified the run/sweep dir rewiring by actually launching the real config (single-run and `-m` sweep) locally and inspecting the resulting dirs + `resolved_config.yaml`. **No correctness bugs found** — the change is sound.

Checks performed:

- **LOW — clean** `${hydra:runtime.output_dir}` resolution: single-run resolves `trainer.default_root_dir` (and the dependent `dirpath`/`save_dir`) to `<output_dir>/<date>/<time>`; sweep jobs resolve to `<output_dir>/<date>/<time>/<num>_<override_dirname>`. Confirmed by running both.
- **LOW — clean** `${now:}` timestamp consistency (`config.yaml:14,16`): both jobs of a `-m` launch landed under one shared `<date>/<time>`, so `run.dir` and `sweep.dir` use the same per-launch timestamp. No collision.
- **LOW — clean** `${hydra.job.num}` prefix (`config.yaml:17`): sweep subdirs came out `0_...`, `1_...`; the prefix guarantees uniqueness even if two jobs shared an override set, as the comment claims.
- **LOW — clean** Missing `output_dir` (`???`) path: a launch without `output_dir=` fails fast with `MissingMandatoryValue ... full_key: hydra.run.dir` before `main()` runs. The explicit `validate_training_config` guard (`train.py:238`) is a correct backstop for non-Hydra/test callers, matching its comment.
- **LOW — clean** `resolved_config.yaml` portability (`train.py:308`): `save_resolved_config` resolves `${hydra:runtime.output_dir}` to a **literal absolute path** in the written file (verified), so downstream `model_loader.setup_model` reads a concrete `default_root_dir`, not an unresolvable `${hydra:...}`.
- **LOW — clean** Resume path: the saved `config.yaml` keeps the unresolved `${hydra:runtime.output_dir}`, which would be unresolvable standalone — but `validate_resume_directory` re-resolves it *inside* the active Hydra context (resume runs through `main()`), so a pinned `hydra.run.dir=... do_resume=true` passes config validation (verified to proceed past resume-check to data loading).
- **LOW — clean** `_demo_train.yaml` keeps the B-style `default_root_dir: ${output_dir}` with no hydra block; both `validate_training_config` guards (`output_dir`, `trainer.default_root_dir`) are satisfied. Integration suites untouched by the new uniqueness machinery, which is why `test_train_hydra_dirs.py` exercises the real config — sound coverage decision.
- **LOW — clean** No dropped/renamed symbol left dangling: `grep` for `run_id`/`_RUN_ID` finds only `model_run_dir` (unrelated) and the deleted test. Eval/doc paths updated to drop the obsolete `outputs/` segment; no stray `outputs/` references remain in code.

Tests: `tests/test_train_hydra_dirs.py` + `tests/test_env_validation.py` — **10 passed** locally.

Minor non-blocking note (cosmetic, not a bug): when a sweep passes `trainer.logger=false` on the CLI (as the new test does), `logger=False` leaks into the `override_dirname` subdir name since it is not in `exclude_keys`. Harmless for real sweeps and the test does not depend on it; flagging only for awareness.

PR #256, opened 2026-07-02
Fix build_index skipping index rebuild on num_queries=0 rerun

payalchandak:
## Summary
- `build_index` ran the `_index/` `rmtree` *after* the `n_queries == 0` early return, so a `num_queries=0` rerun following a real run left stale index partitions on disk instead of wiping them.
- Moves the `rmtree` to run before that early return (but still after the `contexts.height != expected` validation, so a bad-shape call raises before touching the filesystem).
- Adds a regression test that builds a real index, then reruns with `num_queries=0` and asserts `_index/` is fully wiped.

Fixes finding #1 from #255.

## Test plan
- [x] `uv run pytest tests/sampler/test_stage3_build_index.py` — 14 passed (incl. new regression test)
- [x] `uv run pytest tests/sampler tests/test_sampler_dataset_integration.py` — 115 passed
- [x] `uv run ruff check` / `uv run ruff format --check` — clean
- [ ] Manual: run the pipeline once with queries, then rerun with `num_queries=0`; confirm the run completes with no `_validate_row_count` mismatch and `_index/` ends up empty

### DIFF
--- a/src/every_query/generate_tasks/sample_tasks.py
+++ b/src/every_query/generate_tasks/sample_tasks.py
@@ -1141,6 +1141,13 @@ def build_index(
             f"len(queries) * num_contexts_per_query ({n_queries} * {num_contexts_per_query} = {expected})"
         )
 
+    # Wipe ``_index/`` before the empty-input early return (not after) so a ``num_queries=0``
+    # rerun following a real run leaves no stale partitions behind (see #255 finding #1); the
+    # early return below must not short-circuit past this cleanup.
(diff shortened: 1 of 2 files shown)


payalchandak:
Closing — the num_queries=0 early-return in build_index is intentional, load-bearing behavior (see Bug #3 / tests/sampler/test_orchestration.py::test_num_queries_zero_writes_empty_dataset from #239), not just a stale-index bug. Revisiting the actual fix approach before reopening.

PR #258, opened 2026-07-02
Document eligibility/ranking/RNG invariants in sample_tasks.py (#255 #2-#4)

payalchandak:
## Summary
Comment-only follow-up to #255, addressing findings #2, #3, #4. No logic changes.

- **#2** — comment explaining why the `n_prediction_times > min` filter must be strict `>`, not `>=` (Stage 2's `rng.integers(low=min, high=n)` draw would be an illegal empty range at `n == min`).
- **#3** — comment at the ranking step (`build_prediction_times`) noting it depends on `_read_prediction_time_shard` already having dropped null-`time` rows, since nulls sort first and would otherwise claim `prediction_time_index = 0`.
- **#4** — reworked the existing Step A comment in `sample_patient_contexts` to lead with an explicit "do NOT replace with `pl.sample`" warning instead of burying it mid-paragraph.

Plan: `docs/plans/2026-07-02-sample-tasks-255-p2-comments.md`

## Test plan
- [x] `uv run pytest tests/sampler/` — 113 passed
- [x] `uv run ruff check src/every_query/generate_tasks/sample_tasks.py` — pass
- [x] `uv run ruff format --check src/every_query/generate_tasks/sample_tasks.py` — pass
- [x] `git diff` confirmed comment-only (no code lines changed)

### DIFF
--- a/src/every_query/generate_tasks/sample_tasks.py
+++ b/src/every_query/generate_tasks/sample_tasks.py
@@ -300,11 +300,11 @@ def sample_patient_contexts(
         )
 
     # Step A — subject indices (consumed first), with replacement.  One row-gather over the table
-    # (same indices for every column) instead of three per-column gathers.
-    # Do NOT replace `rng.integers` with `prediction_time_counts.sample(..., with_replacement=True)`:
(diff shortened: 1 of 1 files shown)


gkondas (approved):
comment changes look good to me

PR #259, opened 2026-07-02
Make QueryDistribution.sample's uniform branch explicit

payalchandak:
## Summary
- Fixes finding #7 from #255: `QueryDistribution.sample`'s duration draw fell through a bare `else` to treat any non-`log-uniform` value as `uniform`, relying entirely on `__post_init__` having already validated `duration_distribution`.
- Makes the branch exhaustive against `_VALID_DISTRIBUTIONS`, raising `AssertionError` if it ever drifts out of sync with `__post_init__`'s validation, instead of silently defaulting.
- Note: `__post_init__` already rejects any missing/blank/invalid `duration_distribution` at construction time (confirmed present at the pinned review commit and covered by an existing test), so this closes a defense-in-depth gap rather than a live silent-default bug.

## Test plan
- [x] `uv run pytest tests/sampler/test_stage1_query_distribution.py -v` — 17/17 pass, including new `test_sample_raises_if_distribution_drifts_from_valid_set`
- [x] `uv run pytest --doctest-modules src/every_query/generate_tasks/sample_tasks.py` — 10/10 pass
- [x] `uv run ruff check` — clean
- [x] `uv run pytest tests/sampler/` — full suite passes, no regressions

### DIFF
--- a/src/every_query/generate_tasks/sample_tasks.py
+++ b/src/every_query/generate_tasks/sample_tasks.py
@@ -168,8 +168,15 @@ class QueryDistribution:
         ``np.random.default_rng(derive_seed(seed, "queries"))``.  Draws happen in a fixed order (codes
         then durations) so output is deterministic for a fixed ``rng``.
 
+        The distribution branch below is exhaustive over ``_VALID_DISTRIBUTIONS``; ``__post_init__``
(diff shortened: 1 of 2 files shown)


gkondas (approved):
Adding the explicit elif branch for uniform makes the function easier to read. lgtm

PR #260, opened 2026-07-02
Fail loudly on empty queries/contexts in build_index

payalchandak:
## Summary

Fixes item #1 from #255.

- `build_index`'s `if n_queries == 0 or contexts.height == 0: return 0` fired *before* the `_index/` `rmtree`, so a `num_queries=0` rerun after a prior real run left the previous run's stale index partitions in place.
- Since there's no supported empty-budget path through the pipeline (Stage 3's "0 rows written" case was never actually exercised outside this edge case), the early return is replaced with two asserts: `assert n_queries > 0` and `assert contexts.height > 0`. The function now fails fast instead of silently no-op'ing.
- Removed `tests/sampler/test_orchestration.py::test_num_queries_zero_writes_empty_dataset`, the regression test asserting `run()` with `num_queries=0` must succeed against an empty dataset — there's no precedent for using the pipeline this way.
- Updated `tests/sampler/test_stage3_build_index.py::test_empty_queries` (renamed to `test_empty_queries_raises`) to assert `build_index([], ...)` now raises `AssertionError` instead of silently succeeding.

## Test plan

- [x] `uv run pytest tests/sampler -q` — 112 passed
- [x] `uv run ruff check` on changed files — all checks passed

---
_Generated by [Claude Code](https://claude.ai/code/session_01VJjSwcqJsjgy2KLMdZiiQu)_

### DIFF
--- a/src/every_query/generate_tasks/sample_tasks.py
+++ b/src/every_query/generate_tasks/sample_tasks.py
@@ -1126,11 +1126,13 @@ def build_index(
         num_contexts_per_query: Number of patient contexts per query (the ``M`` multiplier).
 
     Returns:
-        ``n_shards`` — the number of index partitions written (0 when there are no queries/contexts).
+        ``n_shards`` — the number of index partitions written.
 
     Raises:
         ValueError: If ``contexts.height != len(queries) * num_contexts_per_query``, or if
             any context fails to resolve a prediction time (null after join).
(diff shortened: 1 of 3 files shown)


gkondas (approved):
asserts add good catch to avoid re running bugs, lgtm

PR #269, opened 2026-07-09
Add in-training task-level AUROC tracking via sampled pos/neg pairs

gkondas:
## Summary
This PR introduces a lightweight mechanism for tracking per-task AUROC during training without scoring the entire tuning split. It adds a new pipeline stage (`EQ_sample_task_tracking_pairs`) that pre-computes one positive and one negative example per task, and integrates task-level AUROC computation into the Lightning training loop.

## Key Changes

- **New module: `sample_task_tracking_pairs.py`**
  - Consumes dense per-shard label parquets from `EQ_generate_evaluation_tasks` (typically `split=tuning`)
  - Samples exactly one positive and one negative row per `(query, duration_days)` task
  - Outputs a compact single parquet (2 rows per surviving task) for efficient repeated reads
  - Drops tasks missing either class (AUROC undefined for single-class tasks)
  - Uses deterministic seeding to ensure reproducibility

- **Lightning module enhancements (`lightning_module.py`)**
  - Added optional `task_auroc_tracking` configuration parameter
  - Implemented `_compute_sampled_task_auroc()` method that:
    - Scores the fixed task-tracking dataloader every validation epoch
    - Computes per-task AUROC as a win/tie/loss indicator from one pos/neg pair
    - Macro-averages indicators across tasks to estimate per-task-averaged AUROC
    - Logs `tuning/occurs_auroc_macro_sampled` and task count metrics
  - Added `_move_batch_to_device()` helper for generic tensor field movement
  - Integrated tracking computation into `on_validation_epoch_end()`

- **Configuration and documentation**
  - Added `sample_task_tracking_pairs_config.yaml` with required/optional parameters
  - Updated `train/configs/config.yaml` with commented example of task tracking setup
  - Updated `generate_tasks/README.md` with pipeline diagram and usage examples
  - Registered `EQ_sample_task_tracking_pairs` CLI entry point in `pyproject.toml`

- **Comprehensive test coverage (`test_sample_task_tracking_pairs_cli.py`, `test_lightning_logic.py`)**
  - End-to-end CLI integration test chaining `EQ_generate_evaluation_tasks` → `EQ_sample_task_tracking_pairs`
  - Schema conformance validation
  - Pairing invariant verification (exactly one True + one False per task)
  - Determinism verification across runs
  - Unit tests for sampled AUROC computation with stubbed model
  - Tests for win/tie/loss indicator calculation and macro-averaging
  - Tests for incomplete task filtering and model training mode restoration

## Implementation Details

- **Deterministic sampling**: Uses hash-based ranking within (task, class) groups to avoid positional bias, mirroring the approach in `sample_evaluation_tasks`
- **Unbiased AUROC estimation**: Single-pair AUROC estimate `P(score(pos) > score(neg))` is unbiased but high-variance; macro-averaging across tasks provides a cheap alternative to scoring the full tuning split
- **No-op when unconfigured**: Feature is entirely optional; leaving `task_auroc_tracking` unset disables it without affecting loss-driven checkpointing or early stopping
- **Efficient dataloader**: Built once in `setup()` and reused across validation epochs since tracking labels are offline-generated and fixed

https://claude.ai/code/session_01R9YdMeThQJUujVz6GYK9Fy

### DIFF
--- a/src/every_query/generate_tasks/sample_task_tracking_pairs.py
+++ b/src/every_query/generate_tasks/sample_task_tracking_pairs.py
@@ -158,7 +158,7 @@ def _read_eval_labels(eval_labels_dir: Path, split: str) -> pl.DataFrame:
     parquets = sorted(split_dir.glob("*.parquet"))
     if not parquets:
         raise FileNotFoundError(f"No parquet files found under {split_dir}")
-    return pl.concat([pl.read_parquet(fp) for fp in parquets], how="vertical")
+    return pl.scan_parquet(parquets).collect()
 
 
 def _out_fp(out_dir: Path, split: str) -> Path:
@@ -188,6 +188,12 @@ def run(
     sample_seed = derive_seed(seed, "task_tracking_pairs", split)
     pairs = sample_task_tracking_pairs(labels_df, seed=sample_seed)
     n_tasks = pairs.height // 2
+    if n_tasks == 0:
+        logger.warning(
+            "No task survived tracking-pair sampling (every task lacked a positive or negative "
+            "example); writing an empty file to %s — in-training AUROC tracking will be a no-op.",
+            _out_fp(out_dir, split),
+        )
     logger.info("Sampled %d pos/neg pairs (%d tasks) for tracking.", pairs.height, n_tasks)
 
     aligned = TaskQuerySchema.align(pairs.to_arrow())
--- a/src/every_query/generate_tasks/sample_task_tracking_pairs.py
+++ b/src/every_query/generate_tasks/sample_task_tracking_pairs.py
@@ -1,0 +1,228 @@
+"""Per-task pos/neg pair sampler for cheap in-training AUROC tracking.
+
+Consumes the dense per-shard label parquets written by ``EQ_generate_evaluation_tasks``
+(typically with ``split=tuning``) and, for each ``(query, duration_days)`` task, samples
+exactly one row with ``boolean_value=True`` and one with ``boolean_value=False``.  Tasks
+missing either class are dropped (AUROC is undefined for a single-class task; mirrors
+``evaluate/metrics.py``'s ``_auroc_or_none``).
+
+The output is a single small parquet — two rows per surviving task — meant to be read by
+``EveryQueryLightningModule``'s task-tracking dataloader every validation pass.  Because
+AUROC for a task equals ``P(score(positive) > score(negative))`` for a random positive/
+negative pair, scoring this one pair per task gives an unbiased (if high-variance)
+per-task AUROC estimate; macro-averaging the resulting win/tie/loss indicators across
+tasks estimates macro AUROC at a cost of ``O(n_tasks)`` forward passes instead of
+``O(tuning-split size)``.
+
+Pipeline position:
+``EQ_generate_evaluation_tasks(split=tuning) -> EQ_sample_task_tracking_pairs -> EQ_train``.
+"""
+
+import logging
+from importlib.resources import files
+from pathlib import Path
+
+import hydra
+import polars as pl
+from omegaconf import DictConfig
+
+from every_query.data.schema import TaskQuerySchema
+from every_query.generate_tasks.sample_tasks import _atomic_write_parquet, _require_path_arg
+from every_query.utils.seeds import derive_seed
+
+logger = logging.getLogger(__name__)
+
+
+# ---------------------------------------------------------------------------
+# Pure primitive
+# ---------------------------------------------------------------------------
+
+
+def sample_task_tracking_pairs(labels_df: pl.DataFrame, seed: int) -> pl.DataFrame:
+    """Sample one positive + one negative labeled row per ``(query, duration_days)`` task.
+
+    Args:
+        labels_df: ``TaskQuerySchema``-shaped rows with a non-null-filterable
+            ``boolean_value`` column (rows with a null ``boolean_value``, i.e. censored,
+            are dropped before sampling — the label is undefined there).
+        seed: PRNG seed.  Deterministic in ``(labels_df, seed)``.
+
+    Returns:
+        A frame with the same columns as ``labels_df``, restricted to exactly two rows
+        (one ``True``, one ``False``) per task that had both classes available. Tasks
+        with only one class present are dropped entirely (a warning is logged with the
+        drop count). Sorted by ``(query, duration_days, boolean_value)``.
+
+    Examples:
+        >>> from datetime import datetime
+        >>> df = pl.DataFrame({
+        ...     "subject_id": [1, 2, 3, 4, 5],
+        ...     "prediction_time": [datetime(2024, 1, i) for i in range(1, 6)],
+        ...     "query": ["A", "A", "A", "B", "B"],
+        ...     "duration_days": pl.Series([30.0, 30.0, 30.0, 30.0, 30.0], dtype=pl.Float32),
+        ...     "boolean_value": [True, True, False, True, True],
+        ... })
+        >>> out = sample_task_tracking_pairs(df, seed=0)
+
+        Task ``A`` has both classes, so it survives with exactly 2 rows; task ``B`` is
+        all-positive and is dropped entirely:
+
+        >>> sorted(out["query"].unique().to_list())
+        ['A']
+        >>> out.height
+        2
+        >>> sorted(out["boolean_value"].to_list())
+        [False, True]
+
+        Determinism — same seed, same output:
+
+        >>> a = sample_task_tracking_pairs(df, seed=42)
+        >>> b = sample_task_tracking_pairs(df, seed=42)
+        >>> a.equals(b)
+        True
+
+        Empty input yields an empty frame:
+
+        >>> sample_task_tracking_pairs(df.head(0), seed=0).height
+        0
+    """
+    required = {
+        TaskQuerySchema.subject_id_name,
+        TaskQuerySchema.prediction_time_name,
+        TaskQuerySchema.query_name,
+        TaskQuerySchema.duration_days_name,
+        TaskQuerySchema.boolean_value_name,
+    }
+    missing = required - set(labels_df.columns)
+    if missing:
+        raise ValueError(f"labels_df is missing required column(s) {sorted(missing)}")
+
+    label_col = TaskQuerySchema.boolean_value_name
+    task_cols = [TaskQuerySchema.query_name, TaskQuerySchema.duration_days_name]
+
+    labeled = labels_df.filter(pl.col(label_col).is_not_null())
+    if labeled.height == 0:
+        return labeled
+
+    # Per-(task, class) sample: derive a per-row hash key from
+    # (subject_id, prediction_time, query, duration_days, seed), rank within each
+    # (task, class) group by that key, and keep rank 0 — same deterministic-without-
+    # positional-bias trick as sample_evaluation_tasks.sample_prediction_times_per_subject.
+    ranked = (
+        labeled.with_columns(
+            pl.concat_str(
+                [
+                    pl.col(TaskQuerySchema.subject_id_name).cast(pl.Utf8),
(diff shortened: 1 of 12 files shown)


src/every_query/model/lightning_module.py
```diff
@@ -196,8 +222,84 @@ def _on_epoch_end(self, split: str):
...
+                _, outputs = self.model(batch)
+                # Squeeze only dim 1 (not `outputs.occurs_probs`'s bare `.squeeze()`), so a
+                # trailing partial batch of size 1 doesn't collapse to a 0-d scalar — same
+                # guard `_log_metrics` applies above.
+                probs = outputs.occurs_logits.detach().cpu().squeeze(1).sigmoid().float().tolist()
+                queries = batch.query.detach().cpu().tolist()
+                durations = batch.duration_days.detach().cpu().tolist()
+                labels = batch.occurs.detach().cpu().tolist()
+                for query, duration, label, prob in zip(queries, durations, labels, probs, strict=True):
+                    probs_by_task.setdefault((query, duration), {})[int(label)] = prob
```
gkondas: **Correctness — OOV query codes silently collide and corrupt the tracked AUROC.**

`batch.query` here is the vocab-index encoding from `EveryQueryPytorchDataset.encode_query`, which falls back to `PAD_INDEX` (0) for any query code missing from the tensorized cohort's `code_metadata` vocab (`src/every_query/data/dataset.py:397-402`). Tracking-task query codes are sourced independently of the training vocab (no intersection is enforced anywhere in the pipeline), so two distinct OOV tasks both key as `(0, duration)`: their pos/neg probs overwrite each other in `probs_by_task`, silently skewing `tuning/occurs_auroc_macro_sampled` and `n_tasks` with no warning.

Suggest filtering tracking pairs to in-vocab codes (at sampling or `setup()` time, warning on drops), or grouping by the original string query instead of the encoded index.

src/every_query/model/lightning_module.py
```diff
@@ -158,6 +167,23 @@ def setup(self, stage=None):
...
 
+        if (
+            self.task_tracking_cfg is not None
+            and stage in ("fit", "validate")
+            and self._task_tracking_loader is None
+        ):
+            # Built once — the tracking labels are a fixed, offline-generated file (see
+            # every_query.generate_tasks.sample_task_tracking_pairs), so there's no need to
+            # re-read/re-instantiate it every validation epoch.
+            dataset = EveryQueryPytorchDataset(self.task_tracking_cfg["config"], split=tuning_split)
```
gkondas: **Correctness — split is hardcoded to `tuning` while the sampler's `split` is configurable.**

`EQ_sample_task_tracking_pairs` accepts any `split` and writes to `{out_dir}/{split}/0.parquet`, but this loader always constructs the dataset with `split=tuning_split`. If pairs were sampled for another split, the label rows join against a disjoint subject set inside meds_torchdata and the tracking dataset ends up **silently empty** — the metric just never logs, no error. Either plumb the split through `task_auroc_tracking`, or raise/warn here when the tracking dataset has length 0.

src/every_query/generate_tasks/sample_task_tracking_pairs.py
```diff
@@ -0,0 +1,228 @@
...
+    labels_df = _read_eval_labels(eval_labels_dir, split)
+    logger.info("Loaded %d labeled eval rows from %s", labels_df.height, eval_labels_dir / split)
+
+    sample_seed = derive_seed(seed, "task_tracking_pairs", split)
+    pairs = sample_task_tracking_pairs(labels_df, seed=sample_seed)
+    n_tasks = pairs.height // 2
+    logger.info("Sampled %d pos/neg pairs (%d tasks) for tracking.", pairs.height, n_tasks)
+
+    aligned = TaskQuerySchema.align(pairs.to_arrow())
+    _atomic_write_parquet(pl.from_arrow(aligned), out_fp)
```
gkondas, line 202: **Silent failure — a degenerate 0-task output is written without complaint.**

When every task is dropped for lacking one of the two classes, `run()` still writes an empty `0.parquet` (only an info-level log), and downstream `_compute_sampled_task_auroc` bails on `if not indicators: return` with no log either — so tracking looks enabled (config set, file exists) but the metric never appears for the entire run, with no diagnostic trail. Suggest raising (or at least `logger.warning`) when 0 tasks survive, plus a warning in `setup()` when the tracking dataset is empty.

src/every_query/model/lightning_module.py
```diff
@@ -196,8 +222,84 @@ def _on_epoch_end(self, split: str):
...
+        self.log(
+            "tuning/occurs_auroc_macro_sampled",
+            sum(indicators) / len(indicators),
+            sync_dist=True,
+        )
+        self.log("tuning/occurs_auroc_macro_sampled_n_tasks", float(len(indicators)), sync_dist=True)
+
     def on_validation_epoch_end(self):
         self._on_epoch_end(tuning_split)
+        self._compute_sampled_task_auroc()
```
gkondas: **Efficiency — the tracking pass runs redundantly on every DDP rank.**

There's no rank gate and the hand-built loader never gets a `DistributedSampler`, so with N GPUs every rank scores the identical full pair set each validation epoch, then `sync_dist=True` all-reduces identical scalars. Multi-GPU DDP is a real production mode for this project, so this is N× wasted forward-pass cost.

Careful with the fix: a naive `if self.trainer.is_global_zero:` gate combined with `sync_dist=True` deadlocks (collective mismatch). Either shard the loader across ranks, or compute on rank 0 only and log with `sync_dist=False` / `rank_zero_only=True`.

src/every_query/model/lightning_module.py
```diff
@@ -122,6 +123,7 @@ def __init__(
         model: EveryQueryModel,
         optimizer: Callable[[Iterator[torch.nn.parameter.Parameter]], torch.optim.Optimizer] | None = None,
         LR_scheduler: Callable[[torch.optim.Optimizer], torch.optim.lr_scheduler._LRScheduler] | None = None,
+        task_auroc_tracking: dict[str, Any] | None = None,
```
gkondas: **Config robustness — untyped `dict[str, Any]` fails slow and silently.**

`self.task_tracking_cfg["config"]` raises a bare `KeyError` deep inside `Trainer.fit()`'s setup with no pointer to the yaml, and a typo like `batch_sz` is silently swallowed by `.get("batch_size", 256)`. The sibling inputs in this same constructor (`optimizer`, `LR_scheduler`) go through the hydra `_target_`/`_partial_` convention and fail fast at instantiation time. A tiny dataclass (or `_target_`-instantiated wrapper) for the tracking config would give the same fail-fast behavior.

src/every_query/model/lightning_module.py
```diff
@@ -196,8 +222,84 @@ def _on_epoch_end(self, split: str):
 
             metric.reset()
 
+    @staticmethod
+    def _move_batch_to_device(batch: EveryQueryBatch, device: torch.device) -> EveryQueryBatch:
```
gkondas: **Reuse — this reimplements `lightning.pytorch.utilities.move_data_to_device`.**

The built-in handles plain dataclasses like `EveryQueryBatch` recursively (verified as a drop-in on the installed lightning 2.5.6), uses `non_blocking=True` for non-CPU devices (missing here), and tracks future batch-shape changes for free — the hand-rolled version only walks top-level `torch.Tensor` fields, so a future nested list/dict of tensors would silently stay on CPU. The whole method body can be `return move_data_to_device(batch, device)`.

src/every_query/model/lightning_module.py
```diff
@@ -196,8 +222,84 @@ def _on_epoch_end(self, split: str):
...
+        """
+        updates = {
+            f.name: value.to(device)
+            for f in dataclasses.fields(batch)
+            if isinstance(value := getattr(batch, f.name), torch.Tensor)
+        }
+        return dataclasses.replace(batch, **updates)
+
+    @torch.no_grad()
+    def _compute_sampled_task_auroc(self):
```
gkondas: **Design — this feature fits Lightning's Callback extension point better than the LightningModule.**

The project already wires cross-cutting concerns as hydra-instantiated callbacks in `train/configs/config.yaml` (`trainer.callbacks`: LR monitor, checkpoint, early stopping). A `Callback` with `setup(trainer, pl_module, stage)` / `on_validation_epoch_end(trainer, pl_module)` has access to everything used here (`pl_module.model`, `.device`, `.log`), and nothing checkpoint-related blocks the move — `task_auroc_tracking` is already deliberately excluded from `save_hyperparameters`. As written, each future aux metric will repeat this pattern: another constructor kwarg, another private method, more manual dataloader bookkeeping inside the module.

src/every_query/generate_tasks/sample_task_tracking_pairs.py
```diff
@@ -0,0 +1,228 @@
...
+    label_col = TaskQuerySchema.boolean_value_name
+    task_cols = [TaskQuerySchema.query_name, TaskQuerySchema.duration_days_name]
+
+    labeled = labels_df.filter(pl.col(label_col).is_not_null())
+    if labeled.height == 0:
+        return labeled
+
+    # Per-(task, class) sample: derive a per-row hash key from
+    # (subject_id, prediction_time, query, duration_days, seed), rank within each
+    # (task, class) group by that key, and keep rank 0 — same deterministic-without-
```
gkondas, line 109: **Reuse — same hash→sort→rank→filter sampling trick as `sample_evaluation_tasks.sample_prediction_times_per_subject`, duplicated.**

This comment even names the mirroring. The two sites differ only in parameters (hash cols, group cols, `k`), so a shared helper like `_deterministic_rank_sample(df, hash_cols, group_cols, seed, k)` would cover both without contortion. Otherwise a future fix to the key derivation or hash choice will likely update one copy and miss the other, silently making the two stages sample inconsistently.

src/every_query/model/lightning_module.py
```diff
@@ -196,8 +222,84 @@ def _on_epoch_end(self, split: str):
...
+        estimate of that task's true AUROC (``AUROC = P(score(pos) > score(neg))``).
+        Macro-averaging those indicators across tasks estimates the per-task-averaged AUROC
+        computed offline in ``evaluate/metrics.py`` at a cost of ``O(n_tasks)`` forward
+        examples instead of ``O(tuning-split size)``. No-op if tracking isn't configured.
+        """
+        loader = self._task_tracking_loader
+        if loader is None:
+            return
+
+        was_training = self.model.training
```
gkondas: **Simplification — the eval/train mode toggle is dead code in production.**

The only call site is `on_validation_epoch_end`, where Lightning guarantees the module is still in eval mode (the evaluation loop restores train mode only *after* this hook fires — see `lightning/pytorch/loops/evaluation_loop.py`, `on_run_end`). `was_training` is always `False` in the real loop; the restore branch is exercised only by the unit test that calls this method directly. Consider dropping the toggle (`no_grad` already covers the numerics) or just documenting the eval-mode precondition.

src/every_query/generate_tasks/sample_task_tracking_pairs.py
```diff
@@ -0,0 +1,228 @@
...
+# I/O helpers
+# ---------------------------------------------------------------------------
+
+
+def _read_eval_labels(eval_labels_dir: Path, split: str) -> pl.DataFrame:
+    split_dir = eval_labels_dir / split
+    parquets = sorted(split_dir.glob("*.parquet"))
+    if not parquets:
+        raise FileNotFoundError(f"No parquet files found under {split_dir}")
+    return pl.concat([pl.read_parquet(fp) for fp in parquets], how="vertical")
```
gkondas: **Minor efficiency — eager sequential shard reads.**

`pl.scan_parquet(parquets).collect()` parallelizes I/O and skips materializing N intermediate DataFrames before the concat — and it's already the sibling module's pattern (`sample_tasks.py` uses `pl.scan_parquet(out_files)` for the same union-of-shards read). Modest win for an offline stage, but it's a one-line change.

PR #276, opened 2026-07-15
Extract sampled macro-AUROC into a torchmetrics Metric

gkondas:
## What

`TaskAurocTrackingCallback` did two unrelated jobs: it owned the `DataLoader` over the offline-sampled pos/neg tracking pairs, *and* it hand-rolled metric accumulation in a `probs_by_task: dict[tuple[int, float], dict[int, float]]` dict-of-dicts, pairing rows and computing the macro win/tie/loss estimate inline in `_compute_and_log`.

That second job is what `torchmetrics.Metric` exists for. This PR moves it into a new `SampledMacroAUROC` (`src/every_query/model/sampled_macro_auroc.py`): accumulation becomes `add_state` lists of scored rows, the pairing becomes `compute()`, the per-epoch clear becomes `reset()`, and the OOV `PAD_INDEX` filter becomes a tensor mask in `update()`. The callback keeps the loader and forward pass and drops to a shell around `update`/`compute`/`reset`.

To be precise about what did *not* change: the `probs_by_task` dict-of-dicts still exists, rebuilt from the state lists on each `compute()`. It moved from the callback into the metric; it was not vectorized away. The tracking set is tiny (one pos/neg pair per task) and a Python grouping loop is the clearest thing that works, so that's deliberate — but don't come to this PR looking for a vectorized grouping.

The wins are the state lifecycle (`reset()` at a defined point rather than an ad-hoc dict clear), standalone testability without a fake `log_fn`, cross-batch accumulation, and the OOV mask.

## Why not put the metric on the LightningModule

The suggested reference ([`MEDS_EIC_AR`'s `NextCodeMetrics`](https://github.com/mmcdermott/MEDS_EIC_AR/blob/main/src/MEDS_EIC_AR/training/module.py)) is a `Metric` updated from batches that already flow through `validation_step`. Our tracking pairs don't — they're a separate dataset over a separate `task_labels_dir`, needing their own loader and forward pass. So the reference pattern applies to the accumulate/compute half only.

The loader stays in the callback because `EveryQueryLightningModule.load_from_checkpoint` reconstructs the module from hparams alone and could not supply the tracking dir — a module owning that path would have a metric it structurally cannot populate on restore.

## Behavior

No user-facing change. Class name, hydra `_target_`, and both logged metric names (`tuning/occurs_auroc_macro_sampled`, `..._n_tasks`) are unchanged, so neither `configs/callbacks/task_auroc.yaml` nor `config.yaml` needed an edit, and logged values are identical.

Two details worth a reviewer's eye:

- **`sync_on_compute=False` is load-bearing.** The callback gates scoring on `trainer.is_global_zero` (the pair set is tiny and identical on every rank), so a syncing `compute()` would enter a collective only rank 0 reaches and deadlock. Commented as such in `__init__`. The `dist_reduce_fx="cat"` on each `add_state` never runs as a consequence, and is commented to say so rather than implying working all-gather.
- **`add_state` defaults to `persistent=False`**, so metric state stays out of `state_dict` and existing checkpoints are unaffected.

`compute()` reads its state through torchmetrics' `dim_zero_cat`, which accepts both the list shape (today) and the concatenated-tensor shape a `_sync_dist` would leave behind — so flipping `sync_on_compute` stays a config decision instead of a `compute()`-time crash.

The `metric.reset()` sits in a `finally` so a mid-epoch failure can't leak rows into the next validation pass and silently skew the estimate.

## Tests

Split in two. `TestSampledMacroAUROC` drives the metric directly with tensors — no fake `log_fn` — and adds cases the old suite couldn't reach cheaply: accumulation across batches (a task's pos and neg rows landing in different batches, which is what `compute()`-time pairing buys), `reset()` clearing state, and same-query-different-duration staying distinct tasks. `TestTaskAurocCallbackLogging` keeps a thin check that the shell logs and doesn't leak state across passes; each pass scores a *different* task, so leaked rows surface as `n_tasks == 2`. Mutation-checked: commenting out `metric.reset()` fails it with `assert 2.0 == 1.0`.

Full suite passes locally (`--ignore=tests/training_validity`), plus doctests on the new module and ruff.

**Not verified locally, needs a cluster run:** the multi-GPU DDP path. That's the one thing that would catch a `sync_on_compute` regression, and its failure mode is a deadlock rather than a wrong number.

### DIFF
--- a/src/every_query/model/sampled_macro_auroc.py
+++ b/src/every_query/model/sampled_macro_auroc.py
@@ -12,6 +12,7 @@ This class only owns the accumulate-and-score half.
 
 import torch
 from torchmetrics import Metric
+from torchmetrics.utilities.data import dim_zero_cat
 
 from every_query.data.dataset import EveryQueryBatch
 
@@ -52,6 +53,8 @@ class SampledMacroAUROC(Metric):
         super().__init__(sync_on_compute=False)
         # add_state defaults to persistent=False, so none of this lands in state_dict and
         # existing checkpoints are unaffected.
+        # dist_reduce_fx is declared for shape-correctness only — it never runs, because
+        # sync_on_compute=False above means this metric is never synced across ranks.
         self.add_state("query", default=[], dist_reduce_fx="cat")
         self.add_state("duration", default=[], dist_reduce_fx="cat")
         self.add_state("label", default=[], dist_reduce_fx="cat")
@@ -100,12 +103,14 @@ class SampledMacroAUROC(Metric):
             (1.0, 1)
         """
         probs_by_task: dict[tuple[int, float], dict[int, float]] = {}
-        if self.query:
+        # len() rather than truthiness: after a _sync_dist these states are tensors, not lists,
+        # and `if self.query:` on a multi-element tensor raises.
+        if len(self.query):
             rows = zip(
-                torch.cat(self.query).tolist(),
-                torch.cat(self.duration).tolist(),
-                torch.cat(self.label).tolist(),
-                torch.cat(self.prob).tolist(),
+                dim_zero_cat(self.query).tolist(),
+                dim_zero_cat(self.duration).tolist(),
+                dim_zero_cat(self.label).tolist(),
+                dim_zero_cat(self.prob).tolist(),
                 strict=True,
             )
             for q, duration, label, prob in rows:
--- a/src/every_query/model/sampled_macro_auroc.py
+++ b/src/every_query/model/sampled_macro_auroc.py
@@ -1,0 +1,125 @@
+"""Macro-averaged sampled AUROC as a ``torchmetrics.Metric``.
+
+A task's AUROC equals ``P(score(pos) > score(neg))`` for a random pos/neg pair, so the
+win/tie/loss indicator on a single offline-sampled pair is an unbiased (high-variance)
+estimate of it.  Macro-averaging those indicators across tasks estimates macro AUROC at
+``O(n_tasks)`` forward examples instead of ``O(split size)``.
+
+The pairs are fed in by ``TaskAurocTrackingCallback``, which owns the dataloader over the
+offline-sampled pair set (see ``every_query.generate_tasks.sample_task_tracking_pairs``).
+This class only owns the accumulate-and-score half.
+"""
+
+import torch
+from torchmetrics import Metric
+
+from every_query.data.dataset import EveryQueryBatch
+
+
+class SampledMacroAUROC(Metric):
+    """Accumulate scored pos/neg rows, then macro-average the per-task win/tie/loss indicator.
+
+    A task is keyed by ``(query, duration_days)`` and contributes one indicator only if both a
+    positive and a negative row were seen for it; tasks missing a class are dropped.
+
+    Examples:
+        >>> m = SampledMacroAUROC()
+        >>> query = torch.tensor([10, 10, 20, 20, 30, 30])
+        >>> duration = torch.tensor([7.0] * 6)
+        >>> occurs = torch.tensor([0, 1, 0, 1, 0, 1])
+        >>> probs = torch.tensor([0.1, 0.9, 0.9, 0.1, 0.5, 0.5])
+        >>> m.update(query, duration, occurs, probs)
+        >>> out = m.compute()
+
+        Task 10 wins (1.0), task 20 loses (0.0), task 30 ties (0.5):
+
+        >>> float(out["auroc"]) == (1.0 + 0.0 + 0.5) / 3
+        True
+        >>> int(out["n_tasks"])
+        3
+
+        With nothing usable, ``auroc`` is NaN and callers skip logging on ``n_tasks == 0``:
+
+        >>> out = SampledMacroAUROC().compute()
+        >>> bool(out["auroc"].isnan()), int(out["n_tasks"])
+        (True, 0)
+    """
+
+    def __init__(self):
+        # sync_on_compute=False is load-bearing: the tracking pair set is tiny and identical on
+        # every rank, so TaskAurocTrackingCallback scores it on rank 0 only.  A syncing compute()
+        # would enter a collective that only rank 0 reaches, and deadlock.
+        super().__init__(sync_on_compute=False)
+        # add_state defaults to persistent=False, so none of this lands in state_dict and
+        # existing checkpoints are unaffected.
+        self.add_state("query", default=[], dist_reduce_fx="cat")
+        self.add_state("duration", default=[], dist_reduce_fx="cat")
+        self.add_state("label", default=[], dist_reduce_fx="cat")
+        self.add_state("prob", default=[], dist_reduce_fx="cat")
+
+    def update(
+        self,
+        query: torch.Tensor,
+        duration_days: torch.Tensor,
+        occurs: torch.Tensor,
+        occurs_probs: torch.Tensor,
+    ) -> None:
+        """Record one batch of scored rows, dropping out-of-vocab queries.
+
+        Out-of-vocab query codes all encode to ``PAD_INDEX``, so distinct OOV tasks would
+        otherwise collide onto one key and corrupt the estimate.
+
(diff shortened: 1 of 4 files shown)


tests/test_lightning_logic.py
```diff
@@ -182,35 +269,29 @@ def _run(model, batches):
...
+        cb = TaskAurocTrackingCallback(config=None)
+        cb._loader = [batch]
+        logged = {}
+        log_fn = lambda name, value, **kw: logged.__setitem__(name, value)  # noqa: E731
 
-        logged = self._run(_StubOccursModel(logits), [batch])
+        for _ in range(2):
+            cb._compute_and_log(_StubOccursModel([-10.0, 10.0]), torch.device("cpu"), log_fn)
 
         assert logged["tuning/occurs_auroc_macro_sampled_n_tasks"] == 1.0
```
gkondas, line 328: This test cannot fail for the reason it exists — verified by mutation: commenting out `self.metric.reset()` in `_compute_and_log` leaves it **passing**.

Both passes feed identical rows, so pass 2 overwrites `probs_by_task[(10, 7.0)]` with the same `{0: 0.1, 1: 0.9}` values. `n_tasks` is 1 whether or not state leaked. The assertion is satisfied by the leak it is supposed to catch.

To actually exercise the reset, the second pass needs a *different* task, so a leak would show up as `n_tasks == 2`:

```suggestion
        assert logged["tuning/occurs_auroc_macro_sampled_n_tasks"] == 1.0
        # Second pass scores a *different* task; without reset() the leaked rows from pass 1
        # would still be in state and this would read 2.0.
```

(Requires reworking the loop above to vary the query per pass — e.g. build a fresh `_make_tracking_batch([20, 20], ...)` for the second iteration.)

src/every_query/model/sampled_macro_auroc.py
```diff
@@ -0,0 +1,125 @@
...
+            >>> m.update(
+            ...     torch.tensor([10, 10, 20]), torch.tensor([7.0, 7.0, 7.0]),
+            ...     torch.tensor([0, 1, 1]), torch.tensor([0.1, 0.9, 0.5]),
+            ... )
+            >>> out = m.compute()
+            >>> float(out["auroc"]), int(out["n_tasks"])
+            (1.0, 1)
+        """
+        probs_by_task: dict[tuple[int, float], dict[int, float]] = {}
+        if self.query:
```
gkondas: Reuse + latent break: torchmetrics ships `dim_zero_cat` for exactly this, and hand-rolling it makes `compute()` depend on the state always being a `list`.

After a `_sync_dist`, torchmetrics replaces list states with a single concatenated tensor. `if self.query:` on a tensor then raises (verified):

```
RuntimeError: Boolean value of Tensor with more than one value is ambiguous
```

Today `sync_on_compute=False` keeps states as lists, so this is latent — but that flag is the one knob this class is designed around, and flipping it turns a deliberate config change into a crash at `compute()`. `dim_zero_cat` accepts both shapes:

```python
from torchmetrics.utilities.data import dim_zero_cat
```
```suggestion
        if self.query:
            rows = zip(
                dim_zero_cat(self.query).tolist(),
```
(and the same for `duration` / `label` / `prob` on the following lines).

src/every_query/model/sampled_macro_auroc.py
```diff
@@ -0,0 +1,125 @@
...
+            >>> m = SampledMacroAUROC()
+            >>> m.update(
+            ...     torch.tensor([10, 10, 20]), torch.tensor([7.0, 7.0, 7.0]),
+            ...     torch.tensor([0, 1, 1]), torch.tensor([0.1, 0.9, 0.5]),
+            ... )
+            >>> out = m.compute()
+            >>> float(out["auroc"]), int(out["n_tasks"])
+            (1.0, 1)
+        """
+        probs_by_task: dict[tuple[int, float], dict[int, float]] = {}
```
gkondas: The PR description claims "the dict-of-dicts becomes `add_state` lists" — but the dict-of-dicts is still right here, rebuilt on every `compute()`. It moved from the callback into the metric; it did not go away.

That is a fine outcome (the tracking set is tiny, and a Python grouping loop is the clearest thing that works), but the PR body oversells it. Worth correcting the description so a reviewer is not hunting for a vectorized grouping that was never written.

The real wins this refactor does deliver are `reset()`/state lifecycle, standalone testability, and the OOV mask — the narrative should rest on those.

src/every_query/model/sampled_macro_auroc.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    def __init__(self):
+        # sync_on_compute=False is load-bearing: the tracking pair set is tiny and identical on
+        # every rank, so TaskAurocTrackingCallback scores it on rank 0 only.  A syncing compute()
+        # would enter a collective that only rank 0 reaches, and deadlock.
+        super().__init__(sync_on_compute=False)
+        # add_state defaults to persistent=False, so none of this lands in state_dict and
+        # existing checkpoints are unaffected.
+        self.add_state("query", default=[], dist_reduce_fx="cat")
```
gkondas, line 62: `dist_reduce_fx="cat"` is inert here: with `sync_on_compute=False` (line 51) the reduce function is never invoked, since `_sync_dist` never runs.

It is harmless today but advertises cross-rank aggregation this metric deliberately does not perform — a reader could reasonably conclude DDP all-gather works and build on that assumption. Since the rank-0-only design is load-bearing, worth a note tying the two together, e.g.:

```suggestion
        # dist_reduce_fx is declared for shape-correctness only — it never runs, because
        # sync_on_compute=False above means this metric is never synced across ranks.
        self.add_state("query", default=[], dist_reduce_fx="cat")
```

gkondas:
Review note that could not be attached inline (the file is not in this PR's diff):

**`src/every_query/model/README.md` — the "What lives here" list is now two files stale.**

It enumerates only `model.py` and `lightning_module.py`. `task_auroc_callback.py` (#269) never got added, and this PR adds `sampled_macro_auroc.py` without an entry either.

That README explicitly documents the per-file layout and states the convention that "a config reader should see exactly which file the class lives in" — so the list drifting undercuts its own stated purpose. Adding both bullets here would leave the directory better documented than this PR found it.

Flagging rather than fixing, since it widens a pre-existing gap rather than creating one.

tests/test_lightning_logic.py
gkondas: Fixed in 2ee56c3. Each pass now scores a different task (10, then 20), so leaked rows surface as a distinct task rather than overwriting an identical key. Re-ran your mutation to confirm it now bites: with `self.metric.reset()` commented out the test fails with `assert 2.0 == 1.0`.

src/every_query/model/sampled_macro_auroc.py
gkondas: Fixed in 2ee56c3 — `compute()` now reads state through `dim_zero_cat`. The `if self.query:` guard was a second copy of the same bug (truthiness on a multi-element tensor raises the same way `torch.cat` did), so that became `if len(self.query):`. Verified by simulating what `_sync_dist` leaves behind — replacing the list states with concatenated tensors and calling `compute()` — which now returns 1.0 / 1 task instead of the ambiguous-bool RuntimeError.

src/every_query/model/sampled_macro_auroc.py
gkondas: Corrected the PR description. It now states outright that the `probs_by_task` dict-of-dicts still exists and is rebuilt from the state lists on each `compute()` — it moved into the metric rather than being vectorized away — and rests the case on the state lifecycle, standalone testability, cross-batch accumulation, and the OOV mask instead.

src/every_query/model/sampled_macro_auroc.py
gkondas: Took the suggestion in 2ee56c3 — the comment now ties `dist_reduce_fx` to `sync_on_compute=False` so it reads as shape-declaration rather than working all-gather.

PR #280, opened 2026-07-20
EQ_generate_evaluation_tasks discovers shards itself

gkondas:
Closes #279.

## What

`EQ_generate_evaluation_tasks` now globs `{data_dir}/data/{split}/*.parquet` and processes every shard found in one invocation — shard discovery is the default and only behavior. The `input_shard` config key is removed entirely, so existing `-m input_shard=...` invocations fail loudly as unknown Hydra overrides instead of silently evaluating a partial split.

## How

- `main` discovers shards via `_split_shards` (reused from `sample_tasks.py`) and loops `run_worker` over them; an empty split dir raises `FileNotFoundError`.
- `run_worker` is untouched: seeds already derive per shard from `(seed, split, input_shard)`, so outputs are byte-identical to the old exhaustive sweep, and the per-shard overwrite guard now doubles as resume-on-rerun.
- Docs: root README loses the shard-counting / `input_shard=range(0,N)` / submitit sweep section; package README examples updated.

## Tests

- Existing CLI tests updated (no more `input_shard=0`); end-to-end test now asserts every input shard gets an output parquet.
- New `test_eq_generate_evaluation_tasks_rejects_input_shard_override` locks in the breaking-change contract.
- `uv run pytest tests/test_generate_evaluation_tasks_cli.py tests/test_sample_task_tracking_pairs_cli.py tests/test_cli_smoke.py` — 26 passed.

### DIFF
--- a/src/every_query/generate_tasks/sample_evaluation_tasks.py
+++ b/src/every_query/generate_tasks/sample_evaluation_tasks.py
@@ -167,7 +167,7 @@ def sample_prediction_times_per_subject(
 def build_evaluation_index_df(
     prediction_times: pl.DataFrame,
     codes: list[str],
-    durations: list[int],
+    durations: list[float],
 ) -> pl.DataFrame:
     """Cross-join prediction times with ``(codes x durations)`` into the evaluation grid.
 
@@ -208,8 +208,8 @@ def build_evaluation_index_df(
         raise ValueError("codes must be non-empty")
     if not durations:
         raise ValueError("durations must be non-empty")
-    if any(not isinstance(d, int) for d in durations):
-        raise TypeError(f"durations must all be ints (got {[type(d).__name__ for d in durations]})")
+    if any(isinstance(d, bool) or not isinstance(d, int | float) for d in durations):
+        raise TypeError(f"durations must all be numbers (got {[type(d).__name__ for d in durations]})")
 
     out_schema = {
         TaskQuerySchema.subject_id_name: prediction_times.schema.get(
@@ -309,7 +309,7 @@ def run_worker(
     split: str,
     input_shard: str,
     codes: list[str],
-    durations: list[int],
+    durations: list[float],
     prediction_times_per_subject: int,
     min_context_per_subject: int,
     seed: int,
@@ -433,32 +433,14 @@ def main(cfg: DictConfig) -> None:
     # Either an explicit list or a dir/path (query_codes=$TENSORIZED_COHORT_DIR loads the full universe).
     codes = read_query_codes(cfg.get("query_codes"))
 
-    # Durations must be whole-day ints — silent truncation (e.g. ``0.5 → 0``) would
-    # change the window semantics.  ``TaskQuerySchema.duration_days`` is float32 on
-    # disk, but the CLI contract here is integer days; fractional durations are a
-    # deliberate-future-feature (see #129), not a silent-today feature.
-    durations: list[int] = []
+    durations: list[float] = []
     for i, d in enumerate(cfg.durations):
         if isinstance(d, bool) or not isinstance(d, int | float):
             raise TypeError(f"cfg.durations[{i}] must be a number, got {type(d).__name__}: {d!r}")
-        if not float(d).is_integer():
-            raise ValueError(
-                f"cfg.durations[{i}] must be a whole-day integer, got {d!r}.  "
-                f"Fractional horizons aren't supported by this CLI yet — see #129."
-            )
-        durations.append(int(d))
+        durations.append(float(d))
 
     split = cfg.split
     shards = _split_shards(data_dir, split)
-    for prev_dir in (Path(out_dir) / "eval" / split, Path(out_dir) / "eval_unique" / split):
-        stale = sorted(p.name for p in prev_dir.glob("*.parquet") if p.stem not in shards)
-        if stale:
-            logger.warning(
-                "Outputs in %s match no discovered shard and will be neither rewritten nor "
-                "removed (downstream consumers read every parquet in this dir): %s",
-                prev_dir,
-                stale,
-            )
 
     # Reject booleans up front: Hydra/OmegaConf parses ``subject_subsample_fraction=true``
     # as a Python ``True``, which would otherwise become ``1.0`` and silently disable
--- a/src/every_query/generate_tasks/sample_evaluation_tasks.py
+++ b/src/every_query/generate_tasks/sample_evaluation_tasks.py
@@ -450,10 +450,15 @@ def main(cfg: DictConfig) -> None:
 
     split = cfg.split
     shards = _split_shards(data_dir, split)
-    if not shards:
-        raise FileNotFoundError(
-            f"No shards found under {data_dir / 'data' / split}; expected *.parquet files."
-        )
+    for prev_dir in (Path(out_dir) / "eval" / split, Path(out_dir) / "eval_unique" / split):
+        stale = sorted(p.name for p in prev_dir.glob("*.parquet") if p.stem not in shards)
+        if stale:
+            logger.warning(
+                "Outputs in %s match no discovered shard and will be neither rewritten nor "
+                "removed (downstream consumers read every parquet in this dir): %s",
+                prev_dir,
+                stale,
+            )
 
     # Reject booleans up front: Hydra/OmegaConf parses ``subject_subsample_fraction=true``
     # as a Python ``True``, which would otherwise become ``1.0`` and silently disable
--- a/src/every_query/generate_tasks/sample_evaluation_tasks.py
+++ b/src/every_query/generate_tasks/sample_evaluation_tasks.py
@@ -8,13 +8,14 @@ row shape needed to compute per-``(query, duration_days)`` metrics over a split
 — every `(subject, time)` gets scored on every task the caller asked about.
 
 Pipeline:
-    1. For the chosen input shard, sample up to ``K`` candidate prediction times
-       per subject (any event time at which the subject has accumulated at least
-       ``min_context_per_subject`` prior events).
+    1. For each shard discovered under ``{data_dir}/data/{split}/*.parquet``,
+       sample up to ``K`` candidate prediction times per subject (any event time
+       at which the subject has accumulated at least ``min_context_per_subject``
+       prior events).
     2. Cross-join with the full ``(codes x durations)`` grid.
     3. Label via :func:`every_query.generate_tasks.sample_tasks.evaluate_index_df`
        (single ``join_asof`` across the whole index frame).
-    4. Align to ``TaskQuerySchema`` and write a single parquet per worker.
+    4. Align to ``TaskQuerySchema`` and write one parquet per shard.
 
 Seeding:
     Prediction-time sampling is deterministic in ``(seed, input_shard, split)``.
@@ -38,6 +39,7 @@ from every_query.generate_tasks.sample_tasks import (
     _atomic_write_parquet,
     _read_event_shard,
     _require_path_arg,
+    _split_shards,
     evaluate_index_df,
     read_query_codes,
 )
@@ -409,20 +411,21 @@ CONFIGS = str(files("every_query") / "generate_tasks" / "configs")
 
 @hydra.main(version_base=None, config_path=CONFIGS, config_name="sample_evaluation_tasks_config")
 def main(cfg: DictConfig) -> None:
-    """Produce one evaluation-tasks parquet per (split, input_shard) pair.
+    """Produce one evaluation-tasks parquet per shard of the chosen split.
+
+    Shards are discovered from ``{data_dir}/data/{split}/*.parquet`` and all of
+    them are processed in this one invocation — there is no shard knob, so a
+    partial split can't be evaluated silently.
 
     Required path args mirror ``sample_tasks`` (``data_dir``, ``out_dir``); eval
     outputs land under a new ``eval/`` subdir so training-task parquets and
     evaluation-task parquets don't collide in one directory.
 
-    Usage (single worker):
+    Usage:
         EQ_generate_evaluation_tasks \\
-            split=held_out input_shard=0 \\
+            split=held_out \\
             prediction_times_per_subject=5 \\
             'query_codes=[HR, TEMP]' 'durations=[1, 7, 30]'
-
-    Sweep across shards with
-    ``python -m every_query.generate_tasks.sample_evaluation_tasks -m input_shard=0,1,2,...``.
     """
     data_dir = _require_path_arg(cfg.get("data_dir"), "data_dir")
     out_dir = _require_path_arg(cfg.get("out_dir"), "out_dir")
@@ -446,7 +449,11 @@ def main(cfg: DictConfig) -> None:
         durations.append(int(d))
 
     split = cfg.split
-    input_shard = str(cfg.input_shard)
+    shards = _split_shards(data_dir, split)
+    if not shards:
+        raise FileNotFoundError(
+            f"No shards found under {data_dir / 'data' / split}; expected *.parquet files."
+        )
 
     # Reject booleans up front: Hydra/OmegaConf parses ``subject_subsample_fraction=true``
     # as a Python ``True``, which would otherwise become ``1.0`` and silently disable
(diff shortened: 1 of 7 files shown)


gkondas:
**[correctness review — behavior regression]** Output-equivalence angle: no findings — the self-discovery loop reproduces the old explicit-shard sweep exactly.

Verified:
- **Shard set**: `_split_shards` (`src/every_query/generate_tasks/sample_tasks.py:1015`, same helper the training generator uses since #279) globs `{data_dir}/data/{split}/*.parquet` sorted, so every shard the old `-m input_shard=range(0,N)` sweep could name is processed exactly once; an empty split now raises `FileNotFoundError` (`sample_evaluation_tasks.py:453-456`) instead of silently doing nothing.
- **Shard→output mapping**: `run_worker` and `_labels_fp`/`_unique_fp` (`sample_evaluation_tasks.py:293-298`) are byte-identical to main — outputs still land at `{out_dir}/eval/{split}/{stem}.parquet` and `eval_unique/…`, and the discovered stem equals the string the old CLI passed as `input_shard`.
- **Seeds / order independence**: `derive_seed(seed, "prediction_times"|"subject_subsample", split, input_shard)` (`sample_evaluation_tasks.py:341,350`) is a blake2b hash over strings, keyed per shard — identical to the old `str(cfg.input_shard)` value — and there is no cross-shard or global RNG state, so the in-process loop order cannot change any shard's content.
- **Skip/overwrite semantics**: the per-shard existing-output skip is unchanged, so partial reruns behave like re-running the old sweep.
- **Training-generator parity**: eval discovery matches the #279 training discovery in both glob pattern and empty-split handling; `split` is passed through unchanged in both.

One non-output operational note (not a regression in results): the sequential loop stops at the first failing shard, whereas a submitit sweep ran shards as independent jobs — a rerun resumes via the existing skip-if-exists logic, so final outputs are still identical.

src/every_query/generate_tasks/sample_evaluation_tasks.py
```diff
@@ -446,7 +449,11 @@ def main(cfg: DictConfig) -> None:
         durations.append(int(d))
 
     split = cfg.split
-    input_shard = str(cfg.input_shard)
+    shards = _split_shards(data_dir, split)
+    if not shards:
+        raise FileNotFoundError(
+            f"No shards found under {data_dir / 'data' / split}; expected *.parquet files."
+        )
```
gkondas: **[simplicity review]** This empty-shards guard is a near-verbatim copy of the one in `build_prediction_times` (`sample_tasks.py:1125-1129`), and those two call sites are the only callers of `_split_shards` — both want the same fail-loud behavior. Fold the raise into `_split_shards` itself and this collapses to the bare call:

```suggestion
    shards = _split_shards(data_dir, split)
```

The other half lives on lines this PR does not touch, so apply manually in `src/every_query/generate_tasks/sample_tasks.py`:

`_split_shards` (line 1015) — before:

```python
def _split_shards(path_to_data: Path, split: str) -> list[str]:
    """Discover shard names for a split: the parquet file stems under ``path_to_data/data/{split}``.

    Sorted for deterministic iteration order.
    """
    split_dir = path_to_data / "data" / split
    return sorted(p.stem for p in split_dir.glob("*.parquet"))
```

after:

```python
def _split_shards(path_to_data: Path, split: str) -> list[str]:
    """Discover shard names for a split: the parquet file stems under ``path_to_data/data/{split}``.

    Sorted for deterministic iteration order.  Raises ``FileNotFoundError`` if the
    split has no shards, so no caller can silently process an empty split.
    """
    split_dir = path_to_data / "data" / split
    shards = sorted(p.stem for p in split_dir.glob("*.parquet"))
    if not shards:
        raise FileNotFoundError(f"No shards found under {split_dir}; expected *.parquet files.")
    return shards
```

and the now-redundant guard in `build_prediction_times` (lines 1125-1129) — before:

```python
    shards = _split_shards(path_to_data, split)
    if not shards:
        raise FileNotFoundError(
            f"No shards found under {path_to_data / 'data' / split}; expected {{i}}.parquet files."
        )
```

after:

```python
    shards = _split_shards(path_to_data, split)
```

Net deletion across the two files, and any future caller gets the empty-split protection for free. Otherwise the PR is minimal: it reuses the existing `_split_shards` helper instead of reinventing discovery, `run_worker` is untouched, and the remaining `input_shard` references are the legitimate per-shard loop variable.

tests/test_generate_evaluation_tasks_cli.py
```diff
@@ -183,6 +187,21 @@ def test_eq_generate_evaluation_tasks_subject_subsample_deterministic(
     )
 
 
+def test_eq_generate_evaluation_tasks_rejects_input_shard_override(tmp_path: Path) -> None:
+    """The removed ``input_shard`` knob must fail loudly, not be silently ignored (#279)."""
+    with pytest.raises(RuntimeError, match="input_shard"):
```
gkondas: **[correctness review]** This test is vacuous as written: `run_and_check` (conftest.py:159-163) embeds the full command line in every `RuntimeError` it raises, so `match="input_shard"` matches the echoed `input_shard=0` argument on *any* nonzero exit. If the knob regressed to being silently accepted (e.g. someone re-adds `input_shard:` to the yaml without wiring it), this invocation would still fail — on the empty `tmp_path` data dir via the new `FileNotFoundError` — and the test would pass anyway. Anchor the match to Hydra's rejection text instead (verified locally: the CLI emits `Could not override 'input_shard'.`).

Related corner: `+input_shard=0` (Hydra append form) is still accepted and silently ignored — all shards get processed. Probably fine (standard Hydra), but the README's "passing `input_shard=` now fails" claim only covers the plain form.
```suggestion
    with pytest.raises(RuntimeError, match="Could not override 'input_shard'"):
```

gkondas:
**[correctness review — callers/integration]** `README.md:176` — this section (rewritten by this PR) still documents the YAML-file form of the codes arg with the wrong key: there is no `codes` key in `sample_evaluation_tasks_config.yaml` (the key is `query_codes`, `src/every_query/generate_tasks/configs/sample_evaluation_tasks_config.yaml:36`, read via `cfg.get("query_codes")` at `sample_evaluation_tasks.py:434`), so the documented command fails as an unknown Hydra override — the exact failure mode the new paragraph at README.md:161 advertises for `input_shard`. It also contradicts the paragraph directly above it, which correctly says `query_codes=` accepts a YAML path. Stale since #161; line 176 is outside this PR's diff hunks so I can't attach a clickable suggestion — exact manual fix:

Before (README.md:176):
```bash
EQ_generate_evaluation_tasks codes=/path/to/sampled_codes.yaml …
```

After:
```bash
EQ_generate_evaluation_tasks query_codes=/path/to/sampled_codes.yaml …
```

gkondas:
**[correctness review]** Behavior regression vs. the old explicit-shard interface, flagging to confirm it's intentional: the old `-m input_shard=range(0,N)` sweep supported `hydra/launcher=submitit_slurm`, i.e. one SLURM job per shard in parallel (this was the documented cluster path in README.md). The new loop in `src/every_query/generate_tasks/sample_evaluation_tasks.py:470-485` processes all discovered shards serially in a single process, and since `input_shard=` is now a rejected override there is no remaining supported way to parallelize across shards or to run a single shard (e.g. a quick smoke on one shard, or re-running just a failed shard without waiting for the skip-scan of the others). Per-shard seeding (`derive_seed(seed, ..., split, input_shard)`) is unchanged, so outputs are bit-identical to the old sweep — correctness of the results is fine; the loss is wall-clock (N shards x per-shard cost in one process) and operational granularity on the cluster. If serial-only is the intent, all good — the skip-if-exists logic does make interrupted runs resumable.

src/every_query/generate_tasks/sample_evaluation_tasks.py
gkondas: **[correctness review — discovery logic]** Stale-output hazard: discovery makes the *input* side exhaustive, but nothing reconciles the *output* dir against the discovered shard set. The loop below only iterates currently-discovered shards, and `run_worker` skips existing outputs (lines 332-334) — so if the upstream split is ever re-sharded/renamed (e.g. 16 shards → 8), the orphaned parquets from the old shard names stay in `out_dir/eval/{split}/` forever, even with `overwrite=true` (overwrite only rewrites shards that still exist). Downstream `EQ_predict` concatenates *every* parquet under the tasks dir (`_validate_tasks_dir` rglob at `src/every_query/predict/predict.py:118-134`, only a warning for multiple files), so it would silently score a mixture of fresh and stale shards — likely with duplicated/contradictory rows for subjects that moved shards. The docstring's claim that "a partial split can't be evaluated silently" is only one-directional: a *superset* split still evaluates silently.

Minimal warn-only fix (deleting when `overwrite=true` would be stronger, but a loud warning at least makes the mixture diagnosable):

```suggestion
    shards = _split_shards(data_dir, split)
    if not shards:
        raise FileNotFoundError(
            f"No shards found under {data_dir / 'data' / split}; expected *.parquet files."
        )
    for prev_dir in (Path(out_dir) / "eval" / split, Path(out_dir) / "eval_unique" / split):
        stale = sorted(p.name for p in prev_dir.glob("*.parquet") if p.stem not in shards)
        if stale:
            logger.warning(
                "Outputs in %s match no discovered shard and will be neither rewritten nor "
                "removed (downstream consumers read every parquet in this dir): %s",
                prev_dir,
                stale,
            )
```

gkondas:
**[correctness review — callers/integration]** Caller sweep for `EQ_generate_evaluation_tasks` — no in-repo breakage found:

- `pyproject.toml:76` entrypoint targets `sample_evaluation_tasks:main`, whose signature is unchanged — fine.
- All three in-repo invocation sites are updated in this PR: `tests/test_generate_evaluation_tasks_cli.py` (all three invocations), `tests/test_sample_task_tracking_pairs_cli.py:35` (the tuning-labels prerequisite step), and `tests/test_cli_smoke.py:29` (help-only, unaffected).
- No SLURM/sbatch scripts, Makefiles, or CI workflows in the repo invoke the entrypoint (`.github/workflows/*` only run pytest; the only repo shell script is `env.example.sh`), and removed `input_shard` cannot be silently ignored — Hydra rejects it as an unknown override, covered by the new `test_eq_generate_evaluation_tasks_rejects_input_shard_override`.
- Downstream consumers are layout-compatible: output still lands at `{out_dir}/eval/{split}/{shard}.parquet` / `eval_unique/…`, matching `EQ_sample_task_tracking_pairs` (`eval_labels_dir/{split}/*.parquet`) and the `${oc.env:EVAL_TASKS_DIR}/eval/...` paths in `src/every_query/predict/external_tasks/configs/*.yaml`. Discovery itself reads no env vars (only the required `data_dir` CLI arg), so there is no eval-side analog of the `$TASK_DIR`/`$TRAINING_TASKS_DIR` mismatch.

One heads-up outside the repo: any cluster-side wrapper/sbatch scripts (e.g. array jobs passing `input_shard=$SLURM_ARRAY_TASK_ID`, or `-m hydra/launcher=submitit_slurm input_shard=range(0,N)` per the old README) will now hard-fail on the unknown override and should be reduced to a single plain invocation — fail-loud as intended, but worth updating those scripts when this merges since the repo can't catch them.

gkondas:
**[correctness review — discovery logic]** Hidden files and directories match the discovery glob. `_split_shards` (`src/every_query/generate_tasks/sample_tasks.py:1015-1021`, pre-existing helper but newly adopted as the eval entry point's shard source in this PR) uses `pathlib.Path.glob("*.parquet")`, which — unlike the `glob` module — **does** match dotfiles, and also matches directories. Verified locally: a dir containing `.hidden.parquet`, `0.parquet`, `1.parquet`, and a directory `dirshard.parquet/` yields stems `['.hidden', '0', '1', 'dirshard']`.

Consequences for the new all-shards loop: a stray `._0.parquet` (macOS AppleDouble file, common after copying data from a Mac onto the cluster) or a directory named `*.parquet` in `{data_dir}/data/{split}/` is discovered as a "shard", and since dotfiles sort first, `pl.read_parquet` crashes on it at the top of the loop before any real shard is processed. The repo's own atomic-write temps (`.{name}.parquet.tmp.*`, `sample_tasks.py:659-675`) are safe — they don't end in `.parquet` — but the input dir is written by upstream tooling this repo doesn't control. Failure is loud, not silent, so this is a robustness hardening, not a correctness bug; note it equally affects the training generator's Stage 0, which shares the helper.

Fix is outside this PR's diff so no suggestion block — in `sample_tasks.py:1021`:

```python
# before
    return sorted(p.stem for p in split_dir.glob("*.parquet"))

# after
    return sorted(
        p.stem for p in split_dir.glob("*.parquet") if p.is_file() and not p.name.startswith(".")
    )
```

Other discovery mechanics check out: `sorted()` makes iteration order deterministic across filesystems (and per-shard seeding is order-independent anyway), stems round-trip exactly to `{stem}.parquet` in `run_worker`, stems are unique per directory so no duplicate/overlapping matches, and an empty or missing `{data_dir}/data/{split}` raises `FileNotFoundError` loudly (`sample_evaluation_tasks.py:453-456`) rather than producing silent empty output.

gkondas:
**[verification — applied fixes]** Verified commit cbc20e9 against all three review suggestions — all applied correctly and completely.

1. **Empty-shards guard folded into `_split_shards`** (`sample_tasks.py:1015`): helper now raises `FileNotFoundError` on an empty split, docstring updated. Both former call-site guards deleted (`sample_evaluation_tasks.py` main and `build_prediction_times`), and a grep confirms those two sites are the only callers — no caller left unprotected.
2. **Orphan-output warning** (`sample_evaluation_tasks.py:453-461`): scans `out_dir/eval/{split}` and `out_dir/eval_unique/{split}`, which match exactly where `_labels_fp`/`_unique_fp` write (`out_dir / split / {shard}.parquet` under the `eval`/`eval_unique` roots passed at lines 478/489). `Path.glob` on a not-yet-existing dir yields nothing, so first runs are unaffected; `logger` is module-level. The merge of the two overlapping suggestions is coherent — the empty-split raise now fires inside `_split_shards` before the warning loop runs, so the inline `if not shards: raise` from the second suggestion was correctly dropped rather than duplicated.
3. **Test match string** (`test_generate_evaluation_tasks_cli.py:192`): now `Could not override 'input_shard'` — confirmed genuine Hydra output by running the suite; the test passes against the real CLI, so the match is no longer satisfiable by the echoed command line alone.
4. **README** line 176 now reads `EQ_generate_evaluation_tasks query_codes=...`, matching the actual config key.

Tests: `tests/test_generate_evaluation_tasks_cli.py` 18 passed; `tests/sampler/test_stage0_prediction_times.py` + `tests/sampler/test_stage3_build_index.py` 27 passed. Clean bill — fixes applied correctly, tests pass.

gkondas:
**[final review]** Ready to merge.

Fresh-eyes pass over the full cumulative diff at cbc20e9, plus a caller sweep and test run. The shard-discovery loop is output-identical to the old `-m input_shard` sweep (per-shard `derive_seed` keying unchanged, no cross-shard RNG state, sorted deterministic iteration), and the fix commit is sound on both counts I checked hardest: (1) folding the empty-shards raise into `_split_shards` changes no caller behavior — its only two callers are the eval `main` (which wants the new fail-loud raise) and `build_prediction_times` (whose cache-hit early return precedes the call, so cached runs never scanned shards before either, and `test_no_shards_raises` matches only the exception type, not the reworded message); (2) the orphan-warning loop is safe on first runs (`Path.glob` on a nonexistent dir yields nothing), warning-only, and scans exactly the dirs `_labels_fp`/`_unique_fp` write into. The `input_shard` rejection test asserts on genuine Hydra struct-mode stderr surfaced through `run_and_check`'s RuntimeError. Test run: `tests/test_generate_evaluation_tasks_cli.py`, `tests/test_sample_task_tracking_pairs_cli.py`, and `tests/sampler/` — 142 passed. Two known non-blockers already on record: serial-only execution replaces the submitit per-shard parallelism (intentional per prior thread), and `_split_shards` globbing dotfiles/dirs is a pre-existing robustness note shared with the training generator, out of scope here. One trivial nit, not blocking: the orphan scan also checks `eval_unique/{split}` when `write_unique_prediction_times=false` — harmless since it only warns.

